PR #85, opened 2016-01-12
fixes multiple queries with wrong filter closes #78

### LINKED ISSUE #78: Queries are evaluated twice on failure in some cases
A query will be sent twice when in the interactive shell if the query raises an exception (in the case of an invalid filter).

Example:
```
$ solvebio
>>> Dataset.retrieve('ClinVar/ClinVar').query().filter(review_status_star__in=[])
```

eyalfoni:
fixes issue with queries resulting in multiple queries in cases where filter is formatted incorrectly

### DIFF
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -8,6 +8,7 @@ from .utils.tabulate import tabulate
 
 import copy
 import logging
+from .errors import SolveError
 logger = logging.getLogger('solvebio')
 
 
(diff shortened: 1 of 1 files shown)


solvebio/query.py
```diff
@@ -421,6 +424,11 @@ def _process_filters(self, filters):
         return rv
 
     def __repr__(self):
+        # check that Query object does not have any previos errors
```
davecap: Typo in comment

solvebio/query.py
```diff
@@ -434,6 +442,12 @@ def __getattr__(self, key):
             logger.debug('warmup (__getattr__: %s)' % key)
             self.execute()
 
+        # check that Query object does not have any previos errors
```
davecap: Same typo

PR #95, opened 2016-04-14
Flat CSV Exporter

jercoh:


solvebio/exporters.py
```diff
@@ -200,6 +215,119 @@ def write(self, filename):
             f.close()
 
 
+class FlatCSVExporter(object):
```
davecap: is it possible to subclass `CSVExporter` and use `__init__` and `write()` from the parent class?

jercoh: Sure I'll fix this

solvebio/exporters.py
```diff
@@ -5,18 +5,25 @@
...
     from collections import OrderedDict
 except ImportError:
     # Python 2.6 compatibility
     from ordereddict import OrderedDict
 
+# Fix Python 2.x.
+try:
+    UNICODE_EXISTS = bool(type(unicode))
+except NameError:
+    unicode = lambda s: str(s)
```
davecap: you can try using `six.u('asdfasdf')` instead of `unicode('asdfasdf')`

jercoh:
@davecap pls review

solvebio/exporters.py
```diff
@@ -247,6 +261,98 @@ def write(self, filename):
             f.close()
 
 
+class FlatCSVExporter(CSVExporter):
```
davecap: do you think we should just replace the existing CSV and XLS exporters with this code?

solvebio/exporters.py
```diff
@@ -247,6 +261,98 @@ def write(self, filename):
...
+        try:
+            return str(value)
+        except UnicodeEncodeError:
+            return ''.join(
+                [s.encode('ascii', 'backslashreplace') for s in value]
+            )
+
+    def write(self, filename):
+        if sys.version_info >= (3, 0, 0):
+            f = open(filename, 'w', newline='')
```
davecap: is this newline='' correct? is it necessary?

solvebio/exporters.py
```diff
@@ -247,6 +261,98 @@ def write(self, filename):
...
+        else:
+            f = open(filename, 'wb')
+
+        fieldnames = sorted(list(self.fields))
+        try:
+            writer = csv.DictWriter(f, fieldnames=fieldnames)
+            writer.writerow(dict(zip(fieldnames, fieldnames)))
```
davecap: Is there a way to just get DictWriter to write the header? or is this the correct way?

solvebio/exporters.py
```diff
@@ -247,6 +261,98 @@ def write(self, filename):
...
+            writer.writerows(self.rows)
+        except:
```
davecap: you can just remove the `except: raise`

solvebio/test/test_exports.py
```diff
@@ -20,6 +20,16 @@ def test_csv_exporter(self):
         self.assertTrue(path.isfile('/tmp/test.csv'))
         remove('/tmp/test.csv')
 
+    # FlatCSVExporter
+    def test_flat_csv_exporter(self):
+        dataset = Dataset.retrieve(self.TEST_DATASET_NAME)
+        query = dataset.query()[:10]
+
+        query.export('flat-csv', filename='/tmp/test_flat.csv')
+        self.assertTrue(path.isfile('/tmp/test_flat.csv'))
```
davecap: I guess we might want to test that the flat export is flattening correctly by comparing against a correct file, maybe by MD5 or by content.

solvebio/exporters.py
jercoh: I read this comment here: https://github.com/solvebio/solvebio-python/blob/28cf4e26a36bae2665dfd48ce0398e448546f8ee/solvebio/exporters.py#L252

solvebio/test/test_exports.py
jercoh: yes we could do that for the other tests as well

PR #98, opened 2016-05-11
use expanduser instead of looking for $HOME manually (fixes #97)

### LINKED ISSUE #97: [Windows] Fix login mechanism on Windows ($HOME not set)
On windows we should check a different environment variable for the home directory in the netrc module.

davecap:


davecap:
Windows Server 2012 test successful.

PR #117, opened 2016-12-08
create datasets with a template file or without a template at all

jsh2134:


solvebio/cli/data.py
```diff
@@ -13,38 +14,78 @@ def create_dataset(args):
 
         * dataset (full name)
         * template_id
+        * template_file
+        * no_template
```
davecap: is it necessary to require this or can it be implicit?

solvebio/cli/data.py
```diff
@@ -13,38 +14,78 @@ def create_dataset(args):
...
+            # Validate the template ID
+            try:
+                tpl = solvebio.DatasetTemplate.retrieve(args.template_id)
+            except solvebio.SolveError as e:
+                if e.status_code != 404:
+                    raise e
+                print("No template with ID {0} found!"
+                      .format(args.template_id))
+        elif args.template_file:
+            if args.template_file.rsplit('.', 1)[1] != 'json':
```
davecap: this may fail if there is no extension

solvebio/cli/data.py
```diff
@@ -13,38 +14,78 @@ def create_dataset(args):
...
+                print("Template filetype unsupported. Please pass a json file")
```
davecap: (sorry) JSON is capitalized, also be specific that it needs a .json extension.

PR #125, opened 2017-01-04
support comma-separated genome builds when creating dataset for import

davecap:


davecap:
Doesn't actually work properly, closing in favor of a new PR

Issue #131, opened 2017-02-14
Problem installing solvebio client - curl-config not found


jhuttner:
Stack trace is below:

```
------------------------------------------------------------
/home/vagrant/ENV/bin/pip run on Tue Feb 14 16:03:18 2017
Downloading/unpacking solvebio

  Getting page https://pypi.python.org/simple/solvebio/
  URLs to search for versions for solvebio:
  * https://pypi.python.org/simple/solvebio/
  Analyzing links from page https://pypi.python.org/simple/solvebio/
    Found link https://pypi.python.org/packages/01/a2/82b348a88672752f87999afb2523b31c298cd2170b20215f1e44210d1962/solvebio-1.2.1.tar.gz#md5=58a07aa4dae237f002890a07b22b126d (from https://pypi.python.org/simple/solvebio/), version: 1.2.1
    Found link https://pypi.python.org/packages/04/d1/3cc45db50bdf89cbb329e43adc98a7d382985108a12f09ba409735e48ad6/solvebio-1.9.1.tar.gz#md5=450a9274edb8497ce884b7a8c8eafe57 (from https://pypi.python.org/simple/solvebio/), version: 1.9.1
    Found link https://pypi.python.org/packages/05/ca/60a8ee5ef8d90d7db697e8a561a2f724c06f924831955692796ff78934a5/solvebio-1.9.0.tar.gz#md5=78848b4aebd96d6a813d138815a6ea7f (from https://pypi.python.org/simple/solvebio/), version: 1.9.0
    Found link https://pypi.python.org/packages/06/d7/26d42da2d4e4d0aa9b3162c55ebbd009effd320b3e58fc9c7407a3ec18a6/solvebio-1.13.2.tar.gz#md5=805aef33d564d1559e15a3747955d15b (from https://pypi.python.org/simple/solvebio/), version: 1.13.2
    Found link https://pypi.python.org/packages/07/60/6d5bd4469bb329e05f2a634e8b1b7885d24cc1bc249a0d5cc3cce59a7ac6/solvebio-1.12.0.tar.gz#md5=63578c842a8e86c41b1912ed842f6beb (from https://pypi.python.org/simple/solvebio/), version: 1.12.0
[...]

PR #132, opened 2017-02-14
Provide help message on curl-config error (#131)

jhuttner:


jhuttner:
@davecap, it does not work for Python 3.  Here is the result:

```
setuptools.sandbox.UnpickleableException: ConfigurationError("Could not run curl-config: [Errno 2] No such file or directory: 'curl-config'",)
```

I have pushed a new commit to fix this, and the current state of the branch works again Python 2.7.5 and Python 3.4.3:

```
## With 'libcurl-devel' not installed...

vagrant@localhost (pycurl-fix *) solvebio-python  $ source ~/ENV/bin/activate
(ENV)vagrant@localhost (pycurl-fix *) solvebio-python  $ python --version
Python 2.7.5
(ENV)vagrant@localhost (pycurl-fix *) solvebio-python  $ python setup.py install 2>&1 | tail -n 3

Problem installing pycurl dependency
You probably need to install libcurl-devel (CentOS) or libcurl4-openssl-dev (Ubuntu)
(ENV)vagrant@localhost (pycurl-fix *) solvebio-python  $ deactivate 

vagrant@localhost (pycurl-fix *) solvebio-python  $ source ~/ENV3/bin/activate
(ENV3) vagrant@localhost (pycurl-fix *) solvebio-python  $ python --version
Python 3.4.3
(ENV3) vagrant@localhost (pycurl-fix *) solvebio-python  $ python setup.py install 2>&1 | tail -n 3

Problem installing pycurl dependency
You probably need to install libcurl-devel (CentOS) or libcurl4-openssl-dev (Ubuntu)
(ENV3) vagrant@localhost (pycurl-fix *) solvebio-python  $ deactivate 


## With 'libcurl-devel' installed...

vagrant@localhost (pycurl-fix *) solvebio-python  $ source ~/ENV/bin/activate
(ENV)vagrant@localhost (pycurl-fix *) solvebio-python  $ python --version
Python 2.7.5
(ENV)vagrant@localhost (pycurl-fix *) solvebio-python  $ pip install pycurl
Collecting pycurl
  Using cached pycurl-7.43.0.tar.gz
Installing collected packages: pycurl
  Running setup.py install for pycurl ... done
Successfully installed pycurl-7.43.0
(ENV)vagrant@localhost (pycurl-fix *) solvebio-python  $ deactivate 

vagrant@localhost (pycurl-fix *) solvebio-python  $ source ~/ENV3/bin/activate
(ENV3) vagrant@localhost (pycurl-fix *) solvebio-python  $ python --version
Python 3.4.3
(ENV3) vagrant@localhost (pycurl-fix *) solvebio-python  $ pip install pycurl
Collecting pycurl
  Using cached pycurl-7.43.0.tar.gz
Building wheels for collected packages: pycurl
  Running setup.py bdist_wheel for pycurl ... done
  Stored in directory: /home/vagrant/.cache/pip/wheels/29/c3/75/12eba77a3bd3b8ed0475680f1957d477c4e299081581effd34
Successfully built pycurl
Installing collected packages: pycurl
Successfully installed pycurl-7.43.0
```

setup.py
```diff
@@ -71,7 +71,16 @@
         **extra
     )
 except Exception as e:
-    if isinstance(e.message, str) and 'Could not run curl-config' in e.message:
+    if ((hasattr(e, 'args') and
```
davecap: This seems like an overly complex way to look for a string in an exception...

setup.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
```
davecap: Any idea why these have to be imported on the same line?

jhuttner: I don't think they have to.  I was just attempting to mimic the style of the next line, which does two imports in one shot.

davecap: Oh yea didn't see that!

setup.py
jsh2134: i dunno the answer to this question, a 3 min google was not helpful, but maybe there is a proper way of checking to see if a system package is installed and we include that in here somewhere instead of trying to catch the error message

jsh2134: a pattern more like this: http://stackoverflow.com/questions/11210104/check-if-a-program-exists-from-a-python-script

davecap: also just want to add that `isinstance(x, str)` won't work if x is a unicode object in Python 2, which may not matter for this particular exception anyways.

jsh2134: this looks promising https://pypi.python.org/pypi/pkgconfig/1.2.2

davecap:
Also I just realized that curl is only needed by a part of the codebase that isn't really used. We can probably eliminate this error from ever happening by simply removing that part of the code.

setup.py
jsh2134:
```
ubuntu@ichi:~  $ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13)
>>> import pkgconfig
>>> pkgconfig.exists('glib-2.0')
False
>>> pkgconfig.exists('libcurl')
True
>>> pkgconfig.exists('openssl')
True

```

PR #140, opened 2017-03-14
Fix capacity argument in import dataset command

davecap:
import dataset command also calls the create dataset command when requested.


jsh2134:
added tests for #142 

NOTE: we can't do a hard delete of depositories/datasets from the client so this will leave remnants in the database after every test.

PR #141, opened 2017-03-14
fix issue where template fields were not used in dataset creation

davecap:
also allow dataset creation without a template


jsh2134:
once #140 is merged build off test in there for #142

PR #154, opened 2017-04-18
Add support for DatasetExports

davecap:
Deprecates client-side "Exporters"

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,9 +55,7 @@ try:
         include_package_data=True,
         install_requires=install_requires,
         platforms='any',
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -55,9 +55,7 @@
         include_package_data=True,
         install_requires=install_requires,
         platforms='any',
-        extras_require={
-            'exporters': ['XlsxWriter>=0.8.0', 'pandas>=0.10.0'],
-        },
```
jsh2134: yay!

PR #157, opened 2017-04-21
update follow logic of imports now the commits are decoupled

### LINKED ISSUE #133: Update DatasetImport follow() function to follow DatasetCommits as well
Imports and Commits were decoupled in December-ish. 

https://github.com/solvebio/solvebio-python/blob/52b83331d7448179cc3f5df7277499eee546776f/solvebio/resource/datasetimport.py#L32

Once the import gets to the  `complete` state check for commits and follow() those

jsh2134:
closes #133

### DIFF
--- a/solvebio/resource/datasetimport.py
+++ b/solvebio/resource/datasetimport.py
@@ -36,24 +36,7 @@ class DatasetImport(CreateableAPIResource, ListableAPIResource,
               .format(self.id))
 
         import_status = self.status
-
         while self.status in ['queued', 'running']:
-            if self.status == 'running':
-                for commit in self.dataset_commits:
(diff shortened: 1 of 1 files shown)


solvebio/resource/datasetimport.py
```diff
@@ -66,7 +49,65 @@ def follow(self):
...
+                                  commit.records_total,
+                                  commit.id))
+                else:
+                    print("Commit '{0}' ({1}) is {2}"
+                          .format(commit.title,
+                                  commit.id,
+                                  commit.status))
+
+                # refresh status
+                commit.refresh()
```
davecap: print statuses, then sleep, then refresh all the commits.

Issue #146, opened 2017-03-29
Add support for Expressions and Annotations


davecap:
The new expression evaluate endpoint should have an Expression resource that can be evaluated. Similar to our JS module. It is currently possible to use the low level `client` to hit the endpoint manually.

PR #163, opened 2017-04-26
Add Annotator and Expression support

davecap:
Uses the annotate endpoint which can add fields to lists of records.

Related to #146 

Supports the following usage:

```python
import solvebio
from solvebio import Dataset
solvebio.login()

# Annotate pathogenic records in ClinVar with the number of CIViC records for the variant
query = Dataset("ClinVar/Variants")\
    .query(fields=["variant_sbid", "rs_id", "rcv_accession", "phenotype"])\
    .filter(clinical_significance="pathogenic")\

fields = [
    {
        "name": "civic_count",
        "data_type": "integer",
        "expression": """
            dataset_count(
                'CIViC/Variants',
                entities=[('variant', record.variant_sbid)]
            )
        """
    }
]

for record in query.annotate(fields):
    if record['civic_count'] > 0:
        print record
```

### DIFF
--- a/solvebio/__init__.py
+++ b/solvebio/__init__.py
@@ -137,6 +137,11 @@ def login(**kwargs):
 
     if not (api_key or access_token):
         print('No credentials found. Requests to SolveBio may fail.')
+    else:
+        from solvebio.client import client
+        # Update the client host and token
+        client.set_host()
+        client.set_token()
 
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


davecap:
It's close but the usage isn't quite as good as I want it to be. Maybe we can try it out a bit before we merge it.

davecap:
Might be useful to add `preview()` to the DatasetMigration resource that returns 10 records (without writing them).

davecap:
@jsh2134 please review again when you have a chance.

davecap:
There's actually a bug here, working on a fix.

PR #168, opened 2017-06-01
move commit follow() into commit resource

jsh2134:


solvebio/resource/dataset.py
```diff
@@ -274,3 +278,23 @@ def migrate(self, target, follow=True, **kwargs):
...
+            return active_tasks
+
+        while True:
+            for task in active_tasks:
+                task.follow()
+
+            time.sleep(5.0)
+            active_tasks = self.tasks()
+            if not active_tasks:
+                break
```
davecap: Can you return tasks at the end even if follow mode is on?

solvebio/resource/datasetcommit.py
```diff
@@ -23,3 +25,51 @@ def dataset(self):
...
         from .datasetimport import DatasetImport
         return DatasetImport.retrieve(self['dataset_import_id'])
+
+    def follow(self, loop=True):
+        if not self.is_approved:
+            # Nothing we can do here!
+            print("This commit needs admin approval.")
+            print("Visit the following page to approve them: "
+                  "https://my.solvebio.com/jobs/imports/{0}"
+                  .format(self.id))
```
davecap: Is it possible to approve via code?

jsh2134: It is possible (i think) if you have approval permissions.

Issue #167, opened 2017-05-18
Dataset export of a query requires a limit


dandanxu:
I'm trying to download all of HGNC but with just 2 fields and following the instructions in the new developer docs.

```
In [9]: dataset = Dataset.retrieve('HGNC/3.0.0-2017-03-01/HGNC')

In [10]: query = dataset.query(fields=['gene_symbol', 'name'])

In [11]: export = query.export(format='csv',follow=True)
```

results in
`SolveError: Bad request. {u'limit': [u"'limit' must be between 1 and 1000000"]}`

It seems that like if I can download the entire dataset with just Dataset.retrieve().export, I should be able to do the same with a subset of the fields? 
The download works if I set
`query = dataset.query(fields=['gene_symbol', 'name'], limit=100000)`

jsh2134:
closed with #169

PR #169, opened 2017-06-01
handle max export limit case

jsh2134:
handle case where the default limit for query scrolling gets set to Infinity, while exports upper limit is  a million


jsh2134:
closes #167

PR #172, opened 2017-07-09
Hyrule Support for Python Client

jhuttner:
Not totally complete yet, but most of the way there.

Removes Depository & Depository Version constructs, and adds in Vault and Object.  All tests are passing.

I need to migrate the public datasets in Staging for SolveBio, which will allow me to get rid of the `force_use_v1` flag that I added in here.  (This was done so that I could retrieve the depository-based public datasets (v1), but also create/list/get new-style datasets (v2) since the SolveBio account has the Hyrule flag activated).

Should I auto-create vaults and parent objects (a la `mkdir -p`) if user says `solvebio create-dataset --vault some-new-vault --path /some/new-path new-dataset-name`?


solvebio/cli/main.py
```diff
@@ -93,8 +104,7 @@ class SolveArgumentParser(argparse.ArgumentParser):
                 },
                 {
                     'name': 'dataset',
-                    'help': 'The full name of the dataset '
-                            '(<depository>/<version>/<dataset>)'
+                    'help': 'The name of the dataset'
```
davecap: doesn't `path` include the name of the `dataset`?

solvebio/resource/dataset.py
```diff
@@ -22,75 +22,61 @@ class Dataset(CreateableAPIResource,
...
+        ('id', 'ID'),
+        ('vault_object_name', 'Name'),
         ('description', 'Description'),
     )
 
     @classmethod
-    def get_or_create_by_full_name(cls, full_name, **kwargs):
-        from solvebio import Depository
-        from solvebio import DepositoryVersion
+    def get_by_full_path(cls, full_path, **kwargs):
```
davecap: is there any way to maintain the get_or_create_by_full_name methods in v2?

solvebio/resource/dataset.py
```diff
@@ -99,9 +85,11 @@ def fields(self, name=None, **params):
...
 
         if name:
-            # construct the field's full_name if a field name is provided
-            return DatasetField.retrieve(
-                '/'.join([self['full_name'], name]))
+            fields = client.get(self.fields_url, params)
+            for i in fields['data']:
+                if i['name'] == name:
+                    result = DatasetField.retrieve(i['id'], **params)
+                    return result
```
jsh2134: can we retrieve the field with only a single lookup? using the field name and the dataset id?

solvebio/resource/dataset.py
```diff
@@ -22,75 +22,61 @@ class Dataset(CreateableAPIResource,
               UpdateableAPIResource):
     """
     Datasets are access points to data. Dataset names are unique
-    within versions of a depository.
+    within a vault folder.
     """
+    USES_V2_ENDPOINT = True
```
jsh2134: can we rename this to: `RESOURCE_VERSION = 2` or something? this way we can update these as we add more versions without adding new flags

solvebio/resource/object.py
```diff
@@ -0,0 +1,56 @@
...
+    though more may be added later: folder, file, and SolveBio Dataset.
+    """
+    USES_V2_ENDPOINT = True
+
+    LIST_FIELDS = (
+        ('id', 'ID'),
+        ('description', 'Description'),
+        ('path', 'Path'),
+        ('filename', 'Filename'),
+        ('filename', 'Filename'),
```
jsh2134: dupe

solvebio/resource/object.py
```diff
@@ -0,0 +1,56 @@
...
+        url = cls.class_url()
+        response = client.get(url, params)
+        results = convert_to_solve_object(response)
+        objects = results.data
+
+        if len(objects) > 1:
+            if params.get('allow_multiple'):
+                return objects
+            else:
+                pass
```
jsh2134: return a single object?

jsh2134:
RE: `Should I auto-create vaults and parent objects (a la mkdir -p) if user says solvebio create-dataset --vault some-new-vault --path /some/new-path new-dataset-name?`

I think we should def auto create the objects, but maybe not the vaults? I believe we want to encourage large and organized vaults, but not hundreds of them. although I can see the desire to create the vaults in a single command as well. maybe we want a new flag `--create-vault` ?

solvebio/resource/dataset.py
davecap: Maybe these will be removed before release?

jhuttner: @jsh2134, good idea.  @davecap, these flags will not be removed as they indicate to the client to use the /v2 endpoints.  That other flag, "force_use_v1", will be removed prior to release.

solvebio/cli/main.py
jhuttner: When you get the object back, the path will include the filename, but I was thinking that for the creation, the `--path` value that the user provides would be the parent directory that the dataset will be placed into.  Does this feel awkward, or OK?  Some examples:

```
# Create a dataset at root
solvebio create-dataset --path=/ --name=new-dataset --vault=my-vault

# Create dataset at /ngs-output
solvebio create-dataset --path=/ngs-output --name=new-dataset --vault=my-vault
```

solvebio/test/test_shortcuts.py
```diff
@@ -59,45 +48,40 @@ def _validate_tmpl_fields(self, fields):
...
-                side_effect=fake_dataset_create)
-    @mock.patch('solvebio.resource.DatasetTemplate.create',
-                side_effect=fake_data_tpl_create)
-    def test_create_dataset_upload_template(
-            self,
-            fake_depo_create,
-            fake_depo_version_create,
-            fake_dataset_create,
-            fake_data_tpl_create):
-        dataset_full_name = self.unique_dataset_name()
```
jhuttner: Careful with mock ordering - this can be confusing.  The above mocks are used in the incorrect order.  This wasn't an issue because they were unused, but the correct order is reverse.  Example below.

```
import time
import unittest

import mock


class Tests(unittest.TestCase):

    @mock.patch('time.time', side_effect=lambda: 123)
    @mock.patch('time.asctime', side_effect=lambda: 'that asc time')
    def test_mocks(self, mock1, mock2):
        self.assertEqual(1, 1)

        self.assertEqual(time.time(), 123)
        self.assertEqual(time.asctime(), 'that asc time')

        # The outer decorators wrap inner decorators.  Mocks are passed in
        # to the original function such that the closest mock is furthest left.
        self.assertEqual(mock1(), 'that asc time')
        self.assertEqual(mock2(), 123)


# Recall that the decorated function below:

@make_bold
@make_italic
def hello():
  return "hello world"
  
# Is the same as...

def hello():
  return "hello world"
hello = make_bold(make_italic(hello))
```

solvebio/test/test_shortcuts.py
```diff
@@ -59,45 +48,40 @@ def _validate_tmpl_fields(self, fields):
...
+    @mock.patch('solvebio.resource.Object.all')
+    @mock.patch('solvebio.resource.Dataset.create')
+    @mock.patch('solvebio.resource.DatasetTemplate.create')
+    def test_create_dataset_upload_template(self, TmplCreate,
+                                            DatasetCreate, ObjectAll,
+                                            VaultAll):
+        TmplCreate.side_effect = fake_dataset_tmpl_create
+        DatasetCreate.side_effect = fake_dataset_create
+        ObjectAll.side_effect = fake_object_all
+        VaultAll.side_effect = fake_vault_all
```
jhuttner: I switched to putting the side effect stuff in the function since I think it reads easier.

solvebio/test/test_dataset.py
```diff
@@ -80,8 +81,8 @@ def test_dataset_fields(self):
 | record_type                  | string      |               |
 | refseq_id_ncbi               | string      |               |
 | refseq_ids                   | string      |               |
-| specialist_database_id       | string      |               |
-| specialist_database_links    | string      |               |
+| specialist_database_id       | blob        |               |
+| specialist_database_links    | blob        |               |
```
jhuttner: This was a pre-existing bug, I believe.

solvebio/resource/datasettemplate.py
```diff
@@ -10,9 +10,10 @@ class DatasetTemplate(CreateableAPIResource, ListableAPIResource,
     DatasetTemplates contain the schema of a Dataset, including some
     properties and all the fields.
     """
+    RESOURCE_VERSION = 2
+
     LIST_FIELDS = (
         ('id', 'ID'),
         ('name', 'Name'),
-        ('version', 'Version'),
```
jhuttner: Pre-existing bug.

jhuttner:
Added a `--create-vault` flag per suggestion by @jsh2134.

Removed my temporary force_use_v1 flag since I migrated public datasets in Staging.  

This is ready for final review.  @jsh2134, @davecap, have at it.  

Probably the best thing to do aside from read the code is try out the client.  Just point at staging and use your SolveBio API Key, which you can obtain from inside the UI.

```
export SOLVEBIO_API_KEY="xxxx"
export SOLVEBIO_API_HOST="https://api-stag.solvebio.com"
```

solvebio/resource/dataset.py
jhuttner: Method has been added.

solvebio/resource/object.py
```diff
@@ -0,0 +1,69 @@
...
+    LIST_FIELDS = (
+        ('id', 'ID'),
+        ('object_type', 'Object Type'),
+        ('path', 'Path'),
+        ('filename', 'Filename'),
+        ('description', 'Description'),
+    )
+
+    @classmethod
+    def retrieve_by_full_path(cls, full_path, **params):
```
davecap: How different are these methods? is it that one takes a vault and one doesnt?

solvebio/resource/object.py
```diff
@@ -0,0 +1,69 @@
...
+                    print i
+                raise Exception('Multiple objects found with {} {}'
+                                .format(name, filter_[name]))
+        elif len(objects) == 1:
+            return objects[0]
+        else:
+            raise Exception('Object not found with {} {}'
+                            .format(name, filter_[name]))
+
+    def help(self):
```
davecap: this we can probably remove

solvebio/resource/object.py
```diff
@@ -0,0 +1,69 @@
...
+
+from .solveobject import convert_to_solve_object
+from .apiresource import CreateableAPIResource
+from .apiresource import ListableAPIResource
+from .apiresource import SearchableAPIResource
+from .apiresource import UpdateableAPIResource
+from .apiresource import DeletableAPIResource
+
+
+class Object(CreateableAPIResource,
```
davecap: Does it make sense to add helper methods to Objects, such as children/parent/siblings/etc... to make it easier to traverse vaults?

davecap: Also the following methods could be useful:

Object.copy()
Object.move()
Object.watch()

solvebio/resource/vault.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @classmethod
+    def datasets(cls, partial_path):
+        return cls._list_helper(partial_path, 'dataset')
+
+    @classmethod
+    def objects(cls, partial_path):
+        return cls._list_helper(partial_path, None)
+
+    def help(self):
```
davecap: We can remove this

solvebio/resource/vault.py
```diff
@@ -0,0 +1,70 @@
...
+from ..help import open_help
+
+from .apiresource import CreateableAPIResource
+from .apiresource import ListableAPIResource
+from .apiresource import SearchableAPIResource
+from .apiresource import UpdateableAPIResource
+from .apiresource import DeletableAPIResource
+
+
+class Vault(CreateableAPIResource,
```
davecap: Can all these actions be done?

* Get object in Vault by ID
* Get object by path
* List objects from a search

solvebio/resource/dataset.py
```diff
@@ -22,23 +24,68 @@ class Dataset(CreateableAPIResource,
...
+    @classmethod
+    def get_by_full_path(cls, full_path, **kwargs):
+        from solvebio import Object
+
+        obj = Object.retrieve_by_full_path(full_path)
+        dataset = Dataset.retrieve(obj['dataset_id'], **kwargs)
+        return dataset
+
+    @classmethod
+    def get_or_create_by_name(cls, vault_name, path, name, **kwargs):
```
davecap: Maybe we can just call it get_or_create() for simplicity? We should discuss our recommended convention of passing a full path at all times, or a split path (vault + path + name).

solvebio/resource/dataset.py
```diff
@@ -22,23 +24,68 @@ class Dataset(CreateableAPIResource,
...
+        result = ':'.join([
+            domain,
+            vault_name,
+            os.path.join('/' + path.lstrip('/'), name),
+        ])
+
+        return result
+
+    @classmethod
+    def get_by_full_path(cls, full_path, **kwargs):
```
davecap: We have "retrieve" which works by ID.. and now we have get_by_full_path, should we name them similarly? retrieve_by_path? Can we always require full_path, unless the user is working through a Vault object?

davecap:
How hard would it be to have a recursive uploader command to "sync" a local dir up to a Vault folder?

solvebio/cli/tutorial.md
```diff
@@ -19,31 +19,40 @@ View this tutorial online: https://www.solvebio.com/docs/python-tutorial
...
 
-To list all datasets from all depositories:
+Vaults are like filesystems on a computer.  They can contain files,
+folders, and a special SolveBio-specific object called a Dataset.
+
+To list all datasets from all vaults, run:
 
     Dataset.all()
 
+To retrieve a dataset by its full path, send your account name, the vault
```
jsh2134: not sure how consistent we are with this but `acme` is technically your `account domain`

solvebio/resource/dataset.py
```diff
@@ -24,58 +26,148 @@ class Dataset(CreateableAPIResource,
...
+
+        try:
+            dataset = Dataset.get_by_full_path(test_path)
+
+            # If the dataset exists but the genome_builds don't match,
+            # update it with the new builds.
+            if dataset.is_genomic and \
+                    dataset.genome_builds != kwargs.get('genome_builds'):
+                dataset.genome_builds = kwargs.get('genome_builds')
+                dataset.save()
```
jsh2134: so this will add a new genome build?

jhuttner: I don't really know.  I just copied it from what is presently there:

https://github.com/solvebio/solvebio-python/blob/master/solvebio/resource/dataset.py#L40

Does something need to change with this?

jsh2134: @davecap RIP genome_builds?

davecap: Yes they are deprecated on datasets.

solvebio/resource/vault.py
jhuttner: Done.

solvebio/cli/tutorial.md
jhuttner: Good catch, changed to "account domain".

solvebio/resource/object.py
jhuttner: Done.

solvebio/resource/object.py
jhuttner: When creating a dataset, the user can pass a path, "/foo/bar/baz", and the code recursively checks if all directories exist, creating them along the way if they don't.  That check uses the this retrieve_by_path method, with an additional vault_id parameter passed in in **params.

Retrieve_by_full_path is used when a user attempts to retrieve a dataset using the "account:vault:path" format.  The user-provided value is passed straight through to the API, and the search is performed without any name-to-ID translations on the client.

Basically, one is designed to be used with additional filters, and one is not.

solvebio/resource/vault.py
jhuttner: To get an object by its ID, you must use the Object class:

```
Object.retrieve(488353213969592764)
```

To get an object by its path, again you must use the Object class:

```
Object.all(path='/1000G')
Object.all(path='/1000G', vault_id=7205)
Object.all(path='/1000G', vault_name='public')
```

Search functionality in the API is presently built on top of the Logical Object model.  Accordingly, searching within the client can be performed by the correct `Object.all(query=xxx)` call:

```
Object.all(query='GRC', vault_id=5906)
```

solvebio/resource/dataset.py
```diff
@@ -99,9 +177,11 @@ def fields(self, name=None, **params):
                 'up fields')
 
         if name:
-            # construct the field's full_name if a field name is provided
-            return DatasetField.retrieve(
-                '/'.join([self['full_name'], name]))
+            fields = client.get(self.fields_url, params)
```
jhuttner: Fixed - had to add a name filter to the API to enable this.

solvebio/resource/object.py
davecap: Any way we can consolidate into one function?

solvebio/resource/vault.py
davecap: Looks good, would be great to have helpers from the Vault class to do similar things.

solvebio/resource/object.py
davecap: This is probably a bad idea, but maybe retrieve() can take full_path=X as a kwarg.

solvebio/resource/dataset.py
```diff
@@ -22,23 +24,68 @@ class Dataset(CreateableAPIResource,
...
+        dataset = Dataset.retrieve(obj['dataset_id'], **kwargs)
+        return dataset
+
+    @classmethod
+    def get_or_create_by_name(cls, vault_name, path, name, **kwargs):
+        from solvebio import Vault
+        from solvebio import Object
         from solvebio import SolveError
 
+        create_vault = kwargs.pop('create_vault')
```
jsh2134: errors, if no value is passed. can we add a default value here?

solvebio/resource/dataset.py
```diff
@@ -47,50 +94,81 @@ def get_or_create_by_full_name(cls, full_name, **kwargs):
...
-        return Dataset.create(
-            depository_version_id=version.id,
-            name=dataset_name, title=title, **kwargs)
 
-    def depository_version(self):
-        from .depositoryversion import DepositoryVersion
-        return DepositoryVersion.retrieve(self['depository_version'])
+        vaults = Vault.all(name=vault_name)
+
+        if len(vaults) == 0:
```
jsh2134: change to `len(vaults.objects())` vaults will always be len=5

jhuttner:
@davecap, added a 'solvebio upload' command that can sync a directory.  I have tested it manually extensively, but haven't written any tests for it because the code is contained, and I felt that to test it with mocks would be kind of difficult.

https://github.com/solvebio/solvebio-python/blob/f-hyrule/solvebio/cli/data.py#L109

solvebio/resource/dataset.py
jhuttner: Rename to "get_or_create".

solvebio/resource/dataset.py
jhuttner: `retrieve_by_path` feels too long to me, since then we have `retrieve_or_create_by_path`.  

I renamed these all to `get_by...` and `retrieve` will just be used for ID-based grabbing.

solvebio/resource/object.py
jhuttner: Currently the client seems very focused on I/O operations => get data into SolveBio, take it out for additional analysis.  Doing things in the client that are possible in the UI doesn't feel especially necessary yet.  If you feel strongly that I should add these, I will, but otherwise my vote is to get the Client released in its more-or-less current form, get the customers upgraded, and then come back to the Client based on additional customer feedback and our own internal feedback.

davecap: Yes that seems fine to me. We should create tickets with features and enhancements that we think mirror what you can do in the UI (and more). Lets evaluate the client today and do some QA, test it out a bit.

solvebio/resource/dataset.py
jhuttner: Added, `, False)` as a default.  Good catch.

solvebio/resource/vault.py
jhuttner: Good idea.  I've added a `.objects` property which calls out to the `.all` method with the appropriate `vault_id` filter.  

```
+    vault = Vault.get_personal_vault()
+    vault.objects.all(path='/1000G')
+    vault.objects.all(query='GRC')
```

https://github.com/solvebio/solvebio-python/blob/f-hyrule/solvebio/resource/vault.py#L11

https://github.com/solvebio/solvebio-python/blob/f-hyrule/solvebio/resource/vault.py#L79

solvebio/resource/dataset.py
jhuttner: Fixed in @jsh2134's merge.

solvebio/resource/dataset.py
jhuttner: Added `, False)` default.

PR #174, opened 2017-07-14
force absolute path. fix bug with name lookup

jsh2134:


solvebio/resource/dataset.py
```diff
@@ -73,8 +73,13 @@ def get_or_create(cls, vault_name, path, name, **kwargs):
 
         create_vault = kwargs.pop('create_vault')
 
-        path = re.sub('//+', '/', path)
+        if path[0] != '/':
+            raise Exception(
+                'Paths are absolute and must begin with a "/"'
+                .format(vault_name)
```
davecap: format is unused

PR #180, opened 2017-08-09
update filter creation to correctly handle nested dict filters

jsh2134:
previously sending this filter

```
>>> filters = ~ Filter(ac_score=None, original_ac_score=None)
>>> print filters.filters
 [{'not': {'and': [('original_ac_score', None), ('ac_score', None)]}}]
```

would end up sending the following to the API:  

```
["not","and"]
```


davecap:
Nice catch

PR #186, opened 2017-09-01
Use Object based uploads, deprecate Upload resource

jsh2134:
- create files everywhere using the new v2 Object resource
- remove the Upload resource
- put Manifest uploads in `solvebio:user-vault:/Uploads` location
- expand user object in local path
- fix test to handle vault_id


jsh2134:
this will be a minor version bump since we are removing the Upload resource.

PR #187, opened 2017-09-13
follow migrations should also follow resulting commits

davecap:


jsh2134:
actually....

I was hoping to deprecate these methods in favor of just following the dataset

```
dataset.activity(follow=True)
```

davecap:
Ah ok, that works now?

davecap:
Closed in favor of activity()

PR #191, opened 2017-09-18
Support for isolated SolveClient (with non-global tokens)

davecap:
Ability to pass a custom client (with custom auth credentials) without using global state.


jsh2134:
LGTM... can you add a few tests?

PR #198, opened 2017-10-18
fix issue where task method did not pass client through

davecap:


solvebio/resource/datasetcommit.py
```diff
@@ -22,12 +22,10 @@ class DatasetCommit(CreateableAPIResource, ListableAPIResource,
...
 
     def dataset(self):
-        from .dataset import Dataset
-        return Dataset.retrieve(self['dataset'])
+        self._client.Dataset.retrieve(self['dataset'])
 
     def dataset_import(self):
-        from .datasetimport import DatasetImport
-        return DatasetImport.retrieve(self['dataset_import_id'])
+        self._client.DatasetImport.retrieve(self['dataset_import_id'])
```
jsh2134: this can be updated to the future...see https://github.com/solvebio/api.solvebio.com/blob/master/api/datasets/dataset_tasks.py#L868

```
if self['parent_job_model'] == 'datasets.DatasetImport':  
       self._client.DatasetImport.retrieve(self['parent_job_id'])
```


do we want to add other possible methods in here also?

```
def dataset_migration(self):
    # returns its parent migration
    pass
```

```
def parent_commit(self):
   # if this a rollback, returns its parent 
   pass
```

jsh2134: moved to #199

PR #211, opened 2017-12-04
use the provided app_url instead of flask's url_root (fixes #210)

### LINKED ISSUE #210: Dash auth wrapper redirect URL is incorrect if different from provided app_url
If the app_url is external facing and differs from the url_root known to Flask, an incorrect redirect_uri is created during the auth code flow.

davecap:


jsh2134:
🌮

PR #213, opened 2017-12-13
add DownloadableAPIResource mixin to Object resource and add support for Object download response to the mixin

davecap:


jsh2134:
👍

PR #216, opened 2017-12-20
add reader_params

### LINKED ISSUE #214: [Manifest] Add reader_params to add_url
add reader_params as a kwarg to add_url() so that specific reader configuration parameters can be passed

https://github.com/solvebio/solvebio-python/blob/master/solvebio/resource/manifest.py#L43

jsh2134:
fixes #214

### DIFF
--- a/solvebio/resource/manifest.py
+++ b/solvebio/resource/manifest.py
@@ -24,7 +24,7 @@ class Manifest(object):
     def __init__(self):
         self.manifest = {'files': []}
 
-    def add_file(self, path):
+    def add_file(self, path, **kwargs):
         default_upload_path = solvebio.Vault.get_or_create_uploads_path()
         vault_name = solvebio.Vault.get_personal_vault().name
         print("Uploading file: {0} to {1}".format(path, default_upload_path))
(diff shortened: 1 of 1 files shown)


jsh2134:
actually will not support this right now since the `reader_params` can just be sent with the DatasetImport

jsh2134:
actually this may not be the case. this needs to be confirmed.

jsh2134:
ready for review. API now correctly supports all the manifest parameters, supports pulling them from DatasetImport if not defined in the manifest AND supports overriding them per file in the manifest _should_ someone have a case where they want to do that.

solvebio/resource/manifest.py
```diff
@@ -37,18 +37,16 @@ def add_file(self, path):
...
     def add_url(self, url, **kwargs):
-        self.manifest['files'].append({
-            'url': url,
-            'name': kwargs.get('name'),
-            'format': kwargs.get('format'),
-            'size': kwargs.get('size'),
-            'md5': kwargs.get('md5'),
-            'base64_md5': kwargs.get('base64_md5')
-        })
+        manifest_item = dict(url=url)
```
davecap: Minor point but you `dict(url=url, **kwargs)` should work

PR #264, opened 2018-06-27
Keep Dash scripts.config.serve_locally False

### LINKED ISSUE #263: [Dash] scripts.config.serve_locally is set to True by default when it should be False
The serve_locally setting is set to False by default, but SolveBioDash sets it to True. This can be confusing for apps built with this class that want to load extra JS.

davidhpark:
closes #263

### DIFF
--- a/solvebio/contrib/dash/solvebio_dash.py
+++ b/solvebio/contrib/dash/solvebio_dash.py
@@ -37,7 +37,6 @@ class SolveBioDash(dash.Dash):
 
         super(SolveBioDash, self).__init__(name, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


solvebio/contrib/dash/solvebio_dash.py
```diff
@@ -37,7 +37,6 @@ def __init__(self, name, *args, **kwargs):
 
         super(SolveBioDash, self).__init__(name, *args, **kwargs)
 
-        self.scripts.config.serve_locally = True
         self.config.supress_callback_exceptions = True
```
davecap: might as well remove this one too and keep the default dash setting

PR #268, opened 2018-07-03
Add sentry integration

### LINKED ISSUE #266: [Dash] Add error handler in callback decorator
One enhancement that was brought up is to have some kind of global error handler in Dash. Since all functions in Dash applications are called in one way or another through callbacks, one idea is to implement an error handler in the `app.callback()` decorator.

davidhpark:
closes #266


davecap:
Closing in favor of integrating in the app code instead.

PR #273, opened 2018-07-18
Update Task objects and add shortcut to cancel tasks

jsh2134:
add task.cancel() and also makes Task undateable so that priority can be set

### DIFF
--- a/solvebio/resource/task.py
+++ b/solvebio/resource/task.py
@@ -27,3 +27,15 @@ class Task(ListableAPIResource):
     def follow(self):
         """ Follow the child object but do not loop """
         self.child_object.follow(loop=False)
+
+    def cancel(self):
+        """ Cancel a task """
+        if 'cancel' not in self.available_actions:
+            # It may not be queued/running or another
+            # possible reason.
+            print("Task can not be canceled.")
+            return
+
+        child_ = self.child_object
+        child_.status = 'canceled'
+        child_.save()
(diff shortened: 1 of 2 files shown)


solvebio/resource/task.py
```diff
@@ -27,3 +27,15 @@ def child_object(self):
     def follow(self):
         """ Follow the child object but do not loop """
         self.child_object.follow(loop=False)
+
+    def cancel(self):
```
jsh2134: might actually be cleaner to make Task and UpdateableResource and send the update there.

solvebio/resource/task.py
```diff
@@ -27,3 +27,15 @@ def child_object(self):
...
+        """ Cancel a task """
+        if 'cancel' not in self.available_actions:
```
davecap: The API already validates this and provides a message about the current status as well.

solvebio/resource/task.py
```diff
@@ -27,3 +27,15 @@ def child_object(self):
...
+            # It may not be queued/running or another
+            # possible reason.
+            print("Task can not be canceled.")
+            return
+
+        child_ = self.child_object
```
davecap: Does it have to be the child object? Can't we just set cancel on the task object itself?

davecap: Or is this what you were referring to with your comment above?

davecap:
Add UpdateableAPIResource class to the the Task resource object.

PR #298, opened 2019-03-22
Prevent cookie from expiring immediately

### LINKED ISSUE #296: [DashAuth] Specify cookie `max_age` to prevent immediate expiry
Currently, the `max_age` parameter is not set, resulting in the immediate expiry of the cookies. This forces users to always log in whenever they close their browsers. One solution would be to set the `max_age` to a week in `SolveBioAuth`.

Related: solvebio/www.solvebio.com#1580

davidhpark:
closes #296 

Changes include:
- Default expiry set to 1 week
- Expose `max_age` to allow custom expiry

### DIFF
--- a/solvebio/contrib/dash/solvebio_auth.py
+++ b/solvebio/contrib/dash/solvebio_auth.py
@@ -20,6 +20,7 @@ class SolveBioAuth(OAuthBase):
     AUTH_COOKIE_NAME = 'dash_solvebio_auth'
     TOKEN_COOKIE_NAME = 'solvebio_oauth_token'
 
+    DEFAULT_TOKEN_COOKIE_MAX_AGE = 60 * 60 * 24 * 7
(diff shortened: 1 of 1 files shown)


solvebio/contrib/dash/solvebio_auth.py
```diff
@@ -20,6 +20,7 @@ class SolveBioAuth(OAuthBase):
     AUTH_COOKIE_NAME = 'dash_solvebio_auth'
     TOKEN_COOKIE_NAME = 'solvebio_oauth_token'
 
+    DEFAULT_TOKEN_COOKIE_MAX_AGE = 60 * 60 * 24 * 7
```
davidhpark: Is this obvious that it's 7 days or should I add a comment?

jsh2134: I think this is definitely more obvious

PR #306, opened 2019-08-28
Task Activity improvements

### LINKED ISSUE #285: [Activity] Simplify and cleanup follow() function (allow sleep time to be configured)
When calling `dataset.activity()` or `import.follow()` there is a sleep timeout of between 3-10 seconds depending on the object. This timeout is too long for validating "realtime" activity such as a single record import. 

Make this timeout configurable and remove the redundant code. 

Look in task.py as well as all the other child Task resources.

e.g.

Commits [sleep for 10 seconds](https://github.com/solvebio/solvebio-python/blob/master/solvebio/resource/datasetcommit.py#L57), Exports for [3 seconds](https://github.com/solvebio/solvebio-python/blob/master/solvebio/resource/datasetexport.py#L57)

jsh2134:
- only get 1 Task in activity
- configure limit and sleep time
- sync across Tasks (closes #285)
- add commit follow to migrations

### DIFF
--- a/solvebio/resource/dataset.py
+++ b/solvebio/resource/dataset.py
@@ -343,28 +343,25 @@ class Dataset(CreateableAPIResource,
         """
         statuses = ['running', 'queued', 'pending']
 
-        # NOTE: source_object_id is not being queried here
-        # and therefore active DatasetExports will not appear
-        # in activity
-        activity = Task.all(target_object_id=self.id,
-                            status=','.join(statuses),
-                            limit=limit,
-                            ordering='-updated_at',
-                            client=self._client)
-
-        print("Found {0} active task(s)".format(activity.total))
-        if not follow:
-            return list(activity)
-
         while True:
-            if not activity:
+
+            # NOTE: source_object_id is not being queried here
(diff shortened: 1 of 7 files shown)


solvebio/resource/datasetimport.py
```diff
@@ -51,12 +52,21 @@ def follow(self, loop=True):
                 if self.status == 'running':
                     print("Processing and validating file(s), "
                           "this may take a few minutes...")
+            else:
```
davecap: why not elif 'running', else...

solvebio/resource/dataset.py
```diff
@@ -333,27 +333,40 @@ def migrate(self, target, follow=True, **kwargs):
...
+                            limit=limit,
+                            ordering='-updated_at',
+                            client=self._client)
+
+        print("Found {0} active task(s)".format(activity.total))
         if not follow:
-            return activity
+            return list(activity)
 
         while True:
```
davecap: To remove the recursion here, we could just move the `activity = Task.all()` to the top of the while loop, followed by `if not follow: return list(activity)`

PR #315, opened 2019-10-31
Upload CLI changes, Tag CLI, vault object shortcuts

### LINKED ISSUE #224: [CLI] Make upload shortcut --full-path arg optional?
Currently `--full-path` [is required](https://github.com/solvebio/solvebio-python/blob/master/solvebio/cli/main.py#L186
), although the argument [docstring](https://github.com/solvebio/solvebio-python/blob/master/solvebio/cli/main.py#L188
) says that the default full-path is a user's personal vault. 

Existing:
`solvebio upload --full-path ~/ some/dir/`

Proposed:
`solvebio upload some/dir/`

If we change this, we will need to [change the example in the docs too ](https://docs.solvebio.com/datasets/importing/#command-line-tools)

### LINKED ISSUE #241: [CLI] Add ability to the upload shortcut to create folders if they don't exist
support stuff like: 
`solvebio upload --full-path domain:vault:/newfolder/newfolder/newfolder var/vardict.PASS.txt`

from dave:
maybe there should be a flag like mkdir -p

### LINKED ISSUE #249: Helper shortcuts for navigating in a given folder
right now, something like grabbing all the bed files in a given folder requires some roundabout logic of going down to the vault level and then searching with the folder path

```
folder = Object.retrieve(folder_id)
folder.parent.vault.files(path=folder.path + '.*bed.gz')
folder.vault.files(path=folder.path + '.*bed.gz')
```

it'd be awesome if I could just... Object.retrieve(folder_id).objects() or Object.retrieve(folder_id).files(path='.*csv') or something. it's also more intuitive.

jsh2134:
### `solvebio upload` command changes

- Removes `--path` and `--vault` in favor of `--full-path` (--closes #224)
- Adds `--create-full-path` to create the upload destination if it does not exist (--closes #241)
- Adds `--exclude` parameter to exclude a list of paths (supports unix style globs) from uploading
- Adds `--dry-run` mode to evaluate before running

```
$ solvebio upload --help
usage: solvebio upload [-h] [--version] [--api-host API_HOST]
                       [--api-key API_KEY] [--api-token API_TOKEN]
                       [--full-path FULL_PATH] [--create-full-path]
                       [--exclude EXCLUDE [EXCLUDE ...]] [--dry-run]
                       local_path [local_path ...]

positional arguments:
  local_path            The path to the local file or directory to upload

SolveBio Options:
  -h, --help            show this help message and exit
  --full-path FULL_PATH
                        The full path where the files and folders should be
                        created, defaults to the root of your personal vault
  --create-full-path    Creates --full-path location if it does not exist.
                        NOTE: This will not create new vaults.
  --exclude EXCLUDE [EXCLUDE ...]
                        Paths to files or folder to be excluded from upload.
                        Unix shell-style wildcards are supported.
  --dry-run             Dry run mode will not upload any files or create any
                        folders.
```
### `solvebio tag` command (NEW) 

Adds command to tag objects

```
$ solvebio tag --help
usage: solvebio tag [-h] [--version] [--api-host API_HOST] [--api-key API_KEY]
                    [--api-token API_TOKEN] --tag TAG [--remove]
                    [--exclude EXCLUDE] [--recursive] [--tag-folders-only]
                    [--tag-files-only] [--tag-datasets-only] [--dry-run]
                    [--no-input]
                    full_path [full_path ...]

positional arguments:
  full_path             The full path of the files, folders or datasets to
                        apply the tag updates. Unix shell-style wildcards are
                        supported.

SolveBio Options:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  --api-host API_HOST   Override the default SolveBio API host
  --api-key API_KEY     Manually provide a SolveBio API key
  --api-token API_TOKEN
                        Manually provide a SolveBio OAuth API token
  --tag TAG             A tag to be added/removed. Files, folders and datasets
                        can be tagged. Tags are case insensitive strings.
                        Example tags: --tag GRCh38 --tag Tissue --tag
                        "Foundation Medicine"
  --remove              Will remove tags instead of adding them.
  --exclude EXCLUDE     Paths to files or folder to be excluded from tagging.
                        Unix shell-style wildcards are supported.
  --recursive           Will recursively apply tag updates to all objects
                        within the folders specified.
  --tag-folders-only    Will only apply tags to folders (tags all objects by
                        default).
  --tag-files-only      Will only apply tags to files (tags all objects by
                        default).
  --tag-datasets-only   Will only apply tags to datasets (tags all objects by
                        default).
  --dry-run             Dry run mode will not save tags.
  --no-input            Automatically accept changes (overrides user prompt)
```

### Adds a shortcut to a Vault objects 
Add shortcuts to Object class for retrieving child objects. Can now do: 
```
folder.files()
folder.datasets()
folder.objects()
folder.ls()
````

Previously had to do:
```
folder.vault.files(parent_object_id=folder.id)
folder.vault.datasets(parent_object_id=folder.id)
etc.
```

this closes #249

### Adds recursive shortcut to get all objects within

```
folder.files(recursive=True)
folder.datasets(recursive=True)
folder.objects(recursive=True)
folder.ls(recursive=True)
````

### adds a shortcut to a Dataset object 
if object is a dataset object, you can now do: 
```
object.query()
```
previously the Dataset object had to be retrieved first
```
Dataset.retrieve(object.dataset_id).query()
```

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -344,26 +344,43 @@ def import_file(args):
               .format(mesh_url))
 
 
-def apply_tags(object_, tags, dry_run=False):
+def apply_tags(object_, tags, remove=False, dry_run=False):
 
     def lowercase(x):
         return x.lower()
 
     existing_tags = map(lowercase, object_.tags)
-    new_tags = [tag for tag in tags if lowercase(tag) not in existing_tags]
+    if remove:
+        removal_tags = [tag for tag in tags if lowercase(tag) in existing_tags]
+        if not removal_tags:
+            print('{}Notice: Object {} does not contain any of the following '
+                  'tags: {}'.format('[Dry Run] ' if dry_run else '',
+                                    object_.full_path, ', '.join(tags)))
+            return
 
-    if not new_tags:
-        print('{}Notice: Object {} already contains these tags'
-              .format('[Dry Run] ' if dry_run else '', object_.full_path))
-        return
+        print('{}Notice: Removing tags: {} from object: {}'
+              .format('[Dry Run] ' if dry_run else '',
+                      ', '.join(removal_tags), object_.full_path))
 
-    print('{}Notice: Adding tags: {} to object: {}'
-          .format('[Dry Run] ' if dry_run else '',
-                  ', '.join(new_tags), object_.full_path))
+        object_tags = [tag for tag in existing_tags if tag not in removal_tags]
+        object_.tags = object_tags
 
-    if not dry_run:
+    else:
+        new_tags = [tag for tag in tags if lowercase(tag) not in existing_tags]
+        if not new_tags:
+            print('{}Notice: Object {} already contains these tags: {}'
+                  .format('[Dry Run] ' if dry_run else '',
+                          object_.full_path, ', '.join(tags)))
+            return
+
+        print('{}Notice: Adding tags: {} to object: {}'
+              .format('[Dry Run] ' if dry_run else '',
+                      ', '.join(new_tags), object_.full_path))
+
         object_tags = object_.tags + new_tags
         object_.tags = object_tags
+
+    if not dry_run:
         object_.save()
 
 
@@ -409,7 +426,8 @@ def tag(args):
             should_tag = False
 
         if should_tag:
-            apply_tags(object_, args.tag, dry_run=args.dry_run)
+            apply_tags(object_, args.tag, remove=args.remove,
+                       dry_run=args.dry_run)
         else:
             print("{}WARNING: Excluding {} {} by object_type".format(
                 '[Dry Run] ' if args.dry_run else '',
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -65,11 +65,7 @@ def should_exclude(path, exclude_paths):
 
 
 def _upload_folder(domain, vault, base_remote_path, base_local_path,
-                   local_start, tags, exclude_paths=None,
-                   tag_folders_only=False, tag_files_only=False):
-
-    folder_tags = tags if not tag_files_only else []
-    file_tags = tags if not tag_folders_only else []
+                   local_start, exclude_paths=None):
 
     # Create the upload root folder if it does not exist on the remote
     try:
@@ -81,7 +77,7 @@ def _upload_folder(domain, vault, base_remote_path, base_local_path,
         base_remote_path, path_dict = \
             Object.validate_full_path(base_remote_path)
         base_folder_path = os.path.join(base_remote_path, local_start)
-        _create_folder(vault, base_folder_path, tags=folder_tags)
+        _create_folder(vault, base_folder_path)
 
     # Create folders and upload files
     for abs_local_parent_path, folders, files in os.walk(base_local_path):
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: solvebio/cli/data.py [hunk 1]
```diff
@@ -2,111 +2,125 @@
...
+def should_exclude(path, exclude_paths):
+    if not exclude_paths:
+        return False
+
+    if path in exclude_paths:
+        print("WARNING: Excluding path {} (via --exclude)".format(path))
+        return True
+
+    for exclude_path in exclude_paths:
+        if path.startswith(exclude_path):
```
davecap: If this was changed to use fnmatch would it make things more complicated? `if fnmatch(exclude_path, path) return True`. Checking the prefix is just like always adding a `*` at the end of the exclude path.

solvebio/cli/data.py
```diff
@@ -2,111 +2,125 @@
...
-                       root).lstrip('/'),
-                f,
-            )
+            local_file_path = os.path.join(abs_local_parent_path, f)
+            if should_exclude(local_file_path, exclude_paths):
+                continue
+
             try:
-                Object.get_by_full_path(file_full_path)
+                Object.get_by_full_path(
```
davecap: Should we log that the file already exists?

solvebio/cli/data.py
```diff
@@ -268,3 +302,69 @@ def import_file(args):
...
               .format(mesh_url))
+
+
+def apply_tags(object_, tags, dry_run=False):
+
+    def lowercase(x):
+        return x.lower()
+
+    existing_tags = map(lowercase, object_.tags)
+    new_tags = [tag for tag in tags if lowercase(tag) not in existing_tags]
```
davecap: I think that this logic (tag case) is already handled by the API

solvebio/resource/object.py
```diff
@@ -242,6 +248,50 @@ def upload_file(cls, local_path, remote_path, vault_full_path, **kwargs):
...
+        """Shortcut to query the underlying Dataset object"""
+        from solvebio import Dataset
+
+        if not self.is_dataset:
+            raise SolveError(
+                "The query method can only be used by a dataset. Found a {}"
+                .format(self.object_type))
+
+        # TODO cache _dataset object?
+        _dataset = Dataset.retrieve(id=self.dataset_id)
```
davecap: You can do `Dataset(self.id)` as a shortcut to avoid retrieving it.

solvebio/cli/data.py
jsh2134: It is, however I didn't want to make an API request to the object if there is no change to be made.

davecap: Up to you

solvebio/cli/data.py
jsh2134: switched to using `fnmatch`

solvebio/cli/data.py
jsh2134: ok ill leave it

solvebio/cli/data.py
```diff
@@ -401,18 +414,26 @@ def tag(args):
                 _regex = path_dict['path']
                 vault_id = Vault.get_by_full_path(
                     path_dict['vault_full_path']).id
-            except:
+            except NotFoundError as e:
```
davecap: Catch any exception

PR #317, opened 2019-11-07
clean up `solvebio tag` code

jsh2134:
- move tree traversal and tag() into Object class
- simplify dry_run and confirmation prompt code

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -370,8 +370,25 @@ def tag(args):
 
     objects = []
     for full_path in args.full_path:
-        objects.extend(list(Object.all(glob=full_path)))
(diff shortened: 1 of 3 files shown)


solvebio/cli/data.py
```diff
@@ -345,65 +345,32 @@ def import_file(args):
...
-        object_.tags = object_tags
-
-    if not dry_run:
-        object_.save()
+    if not valid:
+        print("{}WARNING: Excluding {} {} by object_type".format(
+            '[Dry Run] ' if args.dry_run else '',
+            object_.object_type, object_.full_path))
 
     return True
```
davecap: return valid

PR #325, opened 2020-04-03
Added multiple download of files from folder.

PolinaBevad:
### Description

This related to the script that I created to download files from folder on SolveBio as we don't have the ability to download multiple files in interface.

This is not recursive - will download only Files (but perhaps you want this to be recursive by option?).

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -478,7 +478,7 @@ def download(args):
 
     if sb_object.is_file:
         sb_object.download(local_folder_path)
-        print(f'Downloaded: {sb_object.full_path}')
+        print('Downloaded: {}'.format(sb_object.full_path))
 
     elif sb_object.is_folder:
         files = sb_object.files(query=query) if query else sb_object.files()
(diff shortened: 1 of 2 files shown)


solvebio/cli/data.py
```diff
@@ -460,6 +460,41 @@ def import_file(args):
...
+    # If user sets the home directory, expand the path
+    local_folder_path = os.path.expanduser(local_folder_path) if local_folder_path.startswith(
+        '~') else local_folder_path
+
+    if sb_object.is_file:
+        sb_object.download(local_folder_path)
+        print('Downloaded: {}'.format(sb_object.full_path))
+
+    elif sb_object.is_folder:
+        files = sb_object.files(query=query) if query else sb_object.files()
```
jsh2134, line 484: to be consistent with the other commands, maybe we should change this to use the same `glob=` syntax that the `upload` command and `tag` command full_path use?

thoughts @davecap ?

solvebio/cli/main.py
```diff
@@ -260,6 +260,28 @@ class SolveArgumentParser(argparse.ArgumentParser):
...
+                    'flags': '--query',
+                    'help': 'Can be used to filter only needed files from folder (for example, ".vcf").',
+                },
+                {
+                    'name': 'local_path',
+                    'help': 'The path to the local directory where '
+                            'to download files.',
+                }
+            ]
+        },
```
jsh2134, line 284: can you add a `--dry-run` mode like the other commands? Dry run would not download the file or create the local directory, but print that it would (see other commands)

solvebio/cli/data.py
davecap: Yes I think we should use `glob` if possible.

jsh2134:
bundled into #328

PR #328, opened 2020-04-23
CLI to download multiple files

jsh2134:
Builds off of https://github.com/solvebio/solvebio-python/pull/325

```
 $ solvebio download --help
usage: solvebio download [-h] [--version] [--api-host API_HOST]
                         [--api-key API_KEY] [--access-token ACCESS_TOKEN]
                         [--dry-run]
                         full_path local_path

positional arguments:
  local_path            The path to the local directory where to download
                               files.
  full_path      The full path to the files on SolveBio. Supports Unix
                        style globs in order to download multiple files. Note:
                        Downloads are not recursive.
SolveBio Options:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  --api-host API_HOST   Override the default SolveBio API host
  --api-key API_KEY     Manually provide a SolveBio API key
  --access-token ACCESS_TOKEN
                        Manually provide a SolveBio OAuth2 access token
  --dry-run             Dry run mode will not download any files or create any
                        folders.
```

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -478,7 +478,7 @@ def download(args):
 
     if sb_object.is_file:
         sb_object.download(local_folder_path)
-        print(f'Downloaded: {sb_object.full_path}')
+        print('Downloaded: {}'.format(sb_object.full_path))
 
     elif sb_object.is_folder:
         files = sb_object.files(query=query) if query else sb_object.files()
@@ -489,10 +489,10 @@ def download(args):
         for file in files:
             file_obj = Object.retrieve(file.id)
             file_obj.download(local_folder_path)
-            print(f'Downloaded: {file.full_path}')
+            print('Downloaded: {}'.format(file.full_path))
 
     else:
-        print(f'ERROR: {sb_object.full_path} is not file or folder.')
+        print('ERROR: {} is not file or folder.'.format(sb_object.full_path))
 
 
 def should_tag_by_object_type(args, object_):
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -460,6 +460,41 @@ def import_file(args):
     return imports, dataset
 
 
+def download(args):
+    """
+    Given a folder or file, download all the files contained within it (not recursive).
+    """
+    local_folder_path = args.local_path
+    sb_path = args.full_path
+    query = args.query
+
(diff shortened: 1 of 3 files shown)


davecap:
Would it make more sense to have the full path be a required argument to the command?

`solvebio download remote_path local_path`

jsh2134:
I agree with that. Went ahead and made the change. I left the arg name as full_path to make sure its obvious that it is the full_path (or full_path glob) to the object. We could also change it to `remote_full_path`  if you think that is the best name

`solvebio download full_path local_path`

`solvebio download remote_full_path local_path`

davecap:
What happens if you pass a folder name as the full path?

jsh2134:
> What happens if you pass a folder name as the full path?

Ya what do you want to happen? It will download nothing, as it won't find any files. I tired to mitigate this case by [adding a useful error message here](https://github.com/solvebio/solvebio-python/pull/328/files#diff-6fcda87d3c582de192a4643a9d078c28R490)

We don't have any file path glob detection (we could? and if it is a valid full path, use it explicitly?)

davecap:
It also doesn't work recursively yet right? So it won't create folders found in the remote path?

solvebio/cli/data.py
```diff
@@ -460,6 +460,46 @@ def import_file(args):
...
+        print("Creating local download folder {}".format(local_folder_path))
+        if not dry_run:
+            if not os.path.exists(local_folder_path):
+                os.makedirs(local_folder_path)
+
+    # API will determine depth based on number of "/" in the glob
+    # Add */** to match in any vault (recursive)
+    files = Object.all(glob=full_path, limit=1000, object_type='file')
+    if not files:
+        print("No file(s) found at --full-path {}\nIf aattempting to download "
```
davecap, line 490: typo `aatempting`

jsh2134:
> It also doesn't work recursively yet right? So it won't create folders found in the remote path?

Correct, it currently only downloads a items at a single depth. It will only create folders locally if the local_path is a dir

Here is pseudo-output
```
solvebio download vault:/path/to/files/*.txt my/local/path

  Creating my/local/path/ if not exists
  Downloading vault:/path/to/files/file1.txt to my/local/path/file1.txt
  Downloading vault:/path/to/files/file2.txt to my/local/path/file2.txt
```

davecap:
ok!

PR #329, opened 2020-04-24
adds a shortcut to check if an object contains a tag

jsh2134:
Tags are case insensitive. Adds a helper method to the Object resource to check if a resource contains a tag. 

This is a common bit of code that gets written all over the place. 

```
if file.is_tagged("Special"):
      do_special_things()
else:
    pass
```

### DIFF
--- a/solvebio/resource/object.py
+++ b/solvebio/resource/object.py
@@ -354,24 +354,25 @@ class Object(CreateableAPIResource,
     def is_file(self):
         return self.object_type == 'file'
 
-    def tag(self, tags, remove=False, dry_run=False, apply_save=False):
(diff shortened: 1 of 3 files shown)


solvebio/resource/object.py
```diff
@@ -354,33 +354,33 @@ def is_folder(self):
     def is_file(self):
         return self.object_type == 'file'
 
-    def tag(self, tags, remove=False, dry_run=False, apply_save=False):
-        """Add or remove tags on an object"""
+    def is_tagged(self, tag):
```
davecap, line 357: How about `has_tag`? :neckbeard:

PR #331, opened 2020-04-24
Object.get_or_create_by_full_path()

### LINKED ISSUE #287: Give Object instances that represent "datasets" the ability to be queried
It can be confusing that Objects are also Datasets but they don't have the same helper functions (like "query()". If an Object is a type "dataset", it should be possible to at least query it. How can we blend the Dataset class into the Object class?

One way is to have different sub-classes for each Object type, like `DatasetObject`, `FileObject`, `FolderObject`. Or we can deprecate the `Dataset` class itself and replace it with a subclass of Object (called `Dataset`), and add `File` and `Folder` subclasses.

### LINKED ISSUE #326: Convert `Dataset.create` calls to `Object.create` to use newer endpoint
The API now recommends that datasets should be created through the vaults system (i.e. via the `/v2/objects` endpoint instead of `/v1/datasets`). There are two calls to `Dataset.create` and these should be replaced by `Object.create`.

jsh2134:
Allows creating nested paths on the fly, which currently requires way too many lines of code. Takes the existing Dataset.get_or_create_by_full_path() and moves and upgrades it.

Also closes #326
Also closes #287

### DIFF
--- a/solvebio/resource/dataset.py
+++ b/solvebio/resource/dataset.py
@@ -94,9 +94,8 @@ class Dataset(CreateableAPIResource,
 
     def fields(self, name=None, **params):
         if 'fields_url' not in self:
-            raise Exception(
-                'Please use Dataset.retrieve({ID}) before looking '
-                'up fields')
+            # Dataset object may not have been retrieved. Grab it.
+            self.refresh()
 
         if name:
             params.update({
@@ -118,18 +117,16 @@ class Dataset(CreateableAPIResource,
 
     def template(self, **params):
         if 'template_url' not in self:
-            raise Exception(
-                'Please use Dataset.retrieve({ID}) before retrieving '
-                'a template')
+            # Dataset object may not have been retrieved. Grab it.
+            self.refresh()
 
         response = self._client.get(self.template_url, params)
         return convert_to_solve_object(response, client=self._client)
 
     def commits(self, **params):
         if 'commits_url' not in self:
-            raise Exception(
-                'Please use Dataset.retrieve({ID}) before looking '
-                'up commits')
+            # Dataset object may not have been retrieved. Grab it.
+            self.refresh()
 
         response = self._client.get(self.commits_url, params)
         results = convert_to_solve_object(response, client=self._client)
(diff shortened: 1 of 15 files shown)


solvebio/resource/dataset.py
```diff
@@ -39,6 +38,10 @@ class Dataset(CreateableAPIResource,
     def make_full_path(cls, vault_name, path, name, **kwargs):
         from solvebio import SolveError
 
+        print('[Deprecated] Warning this method has been deprecated'
+              'and will be removed in a future released.'
+              'Use self.vault_object.full_path')
```
davecap, line 43: self?

solvebio/resource/object.py
```diff
@@ -316,17 +418,41 @@ def objects(self, **params):
...
+            # helpers
+            'activity', 'saved_queries'
+        ]
 
+        try:
+            return self[name]
+        except KeyError as err:
+            if name in valid_dataset_attrs and self.is_dataset:
+                return getattr(
+                    Dataset(self.dataset_id, client=self._client), name)
```
davecap, line 442: why not getattr(self.dataset, name)

solvebio/resource/object.py
```diff
@@ -316,17 +418,41 @@ def objects(self, **params):
...
+        from solvebio import Dataset
         if not self.is_dataset:
             raise SolveError(
-                "The query method can only be used by a dataset. Found a {}"
+                "Only dataset objects have a Dataset resource. This is a {}"
                 .format(self.object_type))
 
-        return Dataset(self.dataset_id, client=self._client).query(
-            query=query, **params)
+        return Dataset.retrieve(self['dataset_id'], client=self._client)
```
davecap, line 455: what's the difference between self.dataset_id and self['dataset_id']?

solvebio/resource/object.py
jsh2134: we don't need to do the lookup to get the Dataset object in most cases. Wanted to avoid making the GET request if possible, which is why I used the instance.refresh() logic within those methods that needed them

davecap: but then why not just *not* do the lookup in `def dataset`?

PR #338, opened 2020-04-27
Dataset.get_or_create_by_full_path should always return Dataset object

jsh2134:
Fix regression introduced by #331 (for #326)

### DIFF
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -726,9 +726,9 @@ class Query(object):
         from solvebio import DatasetMigration
 
         # Target can be provided as an object or an ID.
-        if hasattr(target, 'id'):
+        try:
             target_id = target.id
-        else:
+        except AttributeError:
             target_id = target
 
         # If a limit is set in the Query and not overridden here, use it.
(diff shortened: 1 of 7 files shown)


solvebio/query.py
```diff
@@ -723,11 +723,10 @@ def migrate(self, target, follow=True, **kwargs):
         * commit_mode
 
         """
-        from solvebio import Dataset
         from solvebio import DatasetMigration
 
-        # Target can be provided as a Dataset, or as an ID.
-        if isinstance(target, Dataset):
+        # Target can be provided as an object or an ID.
+        if hasattr(target, 'id'):
```
davecap, line 729: https://hynek.me/articles/hasattr/

solvebio/resource/dataset.py
```diff
@@ -70,7 +70,8 @@ def get_or_create_by_full_path(cls, full_path, **kwargs):
         # Assert this is a dataset
         kwargs['assert_type'] = 'dataset'
         kwargs['object_type'] = 'dataset'
```
davecap, line 72: why does it need both assert and object_type?

jsh2134: `assert_type` is needed for the `get` in get_or_create and `object_type` is needed for the `create`

PR #343, opened 2020-05-07
handle case where the error message contains a list of dicts

jsh2134:
Failed to print error when the details contained a list of dicts

### DIFF
--- a/solvebio/errors.py
+++ b/solvebio/errors.py
@@ -53,12 +53,8 @@ class SolveError(Exception):
                 else:
                     self.message += '\nError (%s): ' % k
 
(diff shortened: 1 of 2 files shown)


solvebio/errors.py
```diff
@@ -53,7 +53,9 @@ def __init__(self, message=None, response=None):
                 else:
                     self.message += '\nError (%s): ' % k
 
-                if isinstance(v, list):
+                # Join lists of errors by commas, but if
```
davecap, line 56: maybe we should just print v at this point

jsh2134: ya....

PR #345, opened 2020-05-24
Ability to compose a template via subclass

davecap:


solvebio/resource/solveobject.py
```diff
@@ -45,6 +46,9 @@ def __init__(self, id=None, **params):
         if id:
             self[self.ID_ATTR] = id
 
+        for k, v in list(params.items()):
+            self[k] = v
```
jsh2134, line 50: was this moved from somwehere?

PR #353, opened 2020-07-30
Object tagging updates

### LINKED ISSUE #347: object.tag() clear existing tags in solvebio 2.11.0
Hi @jsh2134, @davecap,

I used 2.9.0 previously, so didn't find this before. 
If I try to use object.tag() to remove tags in solvebio-python 2.11.0, all object tags are cleaned up. 

Example of the code:

    obj = dataset.vault_object  # dataset has tags 'current', 'OAPL' etc.
    obj.tag(['current'], remove=True, apply_save=True) 
    print(obj.tags)   # here it will be empty

The reason is here: https://github.com/solvebio/solvebio-python/blob/master/solvebio/resource/object.py#L504
Old tags rewritten by `updated_tags`. Maybe set difference operation can be used here?
Thank you!

nikolamaric:
- fixes tag removal bug (resolves #347 )
- change `<object>.tag(my_tag)` command to save the tag by default (previously `apply_save=False` by default)
- add a shortcut `untag()` that removes a tag 
- allow a single tag in the `tag()` and `untag()` commands  (previously required a list)
- docs available here https://docs.solvebio.com/vaults/basics/#metadata-and-tags

### DIFF
--- a/solvebio/resource/object.py
+++ b/solvebio/resource/object.py
@@ -501,7 +501,8 @@ class Object(CreateableAPIResource,
                       .format('[Dry Run] ' if dry_run else '',
                               ', '.join(removal_tags), self.full_path))
 
-                updated_tags = [tag for tag in tags if not self.has_tag(tag)]
+                tags_for_removal = [tag for tag in tags if self.has_tag(tag)]
+                updated_tags = [tag for tag in self.tags if tag not in tags_for_removal]
             else:
                 print('{}Notice: Object {} does not contain any of the '
                       'following tags: {}'.format(
--- a/solvebio/resource/object.py
+++ b/solvebio/resource/object.py
@@ -1,4 +1,5 @@
 """Solvebio Object API resource"""
+import collections
 import os
 import re
 import base64
@@ -6,6 +7,7 @@ import binascii
 import mimetypes
 
 import requests
+import six
 from requests.packages.urllib3.util.retry import Retry
 
 from solvebio.errors import SolveError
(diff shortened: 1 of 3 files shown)


solvebio/resource/object.py
```diff
@@ -526,3 +528,10 @@ def tag(self, tags, remove=False, dry_run=False, apply_save=False):
             self.save()
 
         return True
+
+    def untag(self, tags, dry_run=False, apply_save=False):
+        """Remove tags on an object"""
+
+        untag_part = partial(self.tag, remove=True)
+
+        return untag_part(tags=tags, dry_run=dry_run, apply_save=apply_save)
```
jsh2134, line 537: Is partial needed? Why not:

```
        return self.tag(tags=tags, remove=True, dry_run=dry_run, apply_save=apply_save)
```

nikolamaric: the habit...it is not needed. Will change to the suggested.

solvebio/resource/object.py
```diff
@@ -491,7 +491,7 @@ def lowercase(x):
 
         return lowercase(tag) in map(lowercase, self.tags)
 
-    def tag(self, tags, remove=False, dry_run=False, apply_save=False):
+    def tag(self, tags, remove=False, dry_run=False, apply_save=True):
```
davecap, line 494: This is not a new issue, but what happens if somebody puts in a string? o.tag("something"), will it add the following tags: [s, o, m, ...]? We might want to add a check to case the input to a list.

solvebio/resource/object.py
```diff
@@ -484,24 +486,34 @@ def is_file(self):
...
             return x.lower()
 
-        return lowercase(tag) in map(lowercase, self.tags)
+        return lowercase(str(tag)) in map(lowercase, self.tags)
 
-    def tag(self, tags, remove=False, dry_run=False, apply_save=False):
+    def tag(self, tags, remove=False, dry_run=False, apply_save=True):
         """Add or remove tags on an object"""
 
+        def iterable(arg):
```
jsh2134, line 499: rename to `is_iterable_non_string` and add a docstring saying its a py2or3 compat way to check

PR #355, opened 2020-08-06
Issue 276 dict attributes not save updates

### LINKED ISSUE #276: Modifying object attributes that are dictionaries may not save updates
If you modify a dictionary on an APIResource and then save() the resource, the updates are not sent. This is because  [__setitem__ is not getting called on the resource](https://github.com/solvebio/solvebio-python/blob/35044c3303e84a8ee7e08a8a8ee6184080b0e668/solvebio/resource/solveobject.py#L63) in this case and not updating the `_unsaved_values` set.

Example
```
# Failure case
print obj.metadata
{}
obj.metadata['foo'] = 'bar'
obj.save()
# nada
print obj.metadata
{}

# Workaround
obj.metadata = {'foo': 'bar'}
obj.save()
print obj.metadata
{'foo': 'bar'}
```

nikolamaric:
closes #276

### DIFF
--- a/solvebio/resource/apiresource.py
+++ b/solvebio/resource/apiresource.py
@@ -312,9 +312,19 @@ class UpdateableAPIResource(APIResource):
 
     def serialize(self, obj):
         params = {}
-        if obj._unsaved_values:
-            for k in obj._unsaved_values:
-                if k == 'id' or k == self.ID_ATTR:
-                    continue
+        unsaved_values = obj._unsaved_values
+
+        # adding 'metadata' attribute to be serialized
+        unsaved_values.add('metadata')
+
+        for k in unsaved_values:
+            if k == 'id' or k == self.ID_ATTR:
+                continue
+
+            attr = getattr(obj, k)
+            if k == 'metadata' and isinstance(attr, SolveObject):
+                params[k] = dict(attr)
+            else:
                 params[k] = getattr(obj, k) or ""
+
         return params
--- a/solvebio/resource/apiresource.py
+++ b/solvebio/resource/apiresource.py
@@ -315,16 +315,18 @@ class UpdateableAPIResource(APIResource):
         unsaved_values = obj._unsaved_values
 
         # adding 'metadata' attribute to be serialized
-        unsaved_values.add('metadata')
+        if getattr(obj, 'metadata'):
+            unsaved_values.add('metadata')
 
         for k in unsaved_values:
             if k == 'id' or k == self.ID_ATTR:
                 continue
 
             attr = getattr(obj, k)
-            if k == 'metadata' and isinstance(attr, SolveObject):
+
+            if isinstance(attr, SolveObject):
                 params[k] = dict(attr)
             else:
-                params[k] = getattr(obj, k) or ""
+                params[k] = attr or ""
 
         return params
--- a/solvebio/test/test_dataset.py
+++ b/solvebio/test/test_dataset.py
@@ -73,21 +73,3 @@ class DatasetTests(SolveBioTestCase):
         field = dataset.fields('status')
         facets = field.facets()
         self.assertTrue(len(facets['facets']) >= 0)
-
-    def test_dataset_metadata(self):
-        dataset = self.client.Dataset.get_by_full_path(
-            self.TEST_DATASET_FULL_PATH)
-        metadata = dataset.metadata
-
-        metadata['foo_1'] = 'bar_1'
-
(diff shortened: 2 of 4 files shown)


solvebio/test/test_dataset.py
```diff
@@ -73,3 +73,21 @@ def test_dataset_facets(self):
...
+        # Test that setting item has been done successfully
+        self.assertTrue(metadata.get('foo_1') == 'bar_1')
+
+        metadata = {'foo_2': 'bar_2'}
+        self.assertTrue(metadata.get('foo_2') == 'bar_2')
+
+        # Test that direct assignement overrides SolveBio object type with the given dictionary
+        self.assertTrue(len(metadata) == 1 and
+                        metadata.pop('foo_2') == 'bar_2' and
+                        len(metadata) == 0)
```
davecap, line 93: Can you add a test for the serialize method that it properly returns the metadata  in params?

solvebio/test/test_dataset.py
```diff
@@ -73,3 +73,21 @@ def test_dataset_facets(self):
         field = dataset.fields('status')
         facets = field.facets()
         self.assertTrue(len(facets['facets']) >= 0)
+
+    def test_dataset_metadata(self):
+        dataset = self.client.Dataset.get_by_full_path(
+            self.TEST_DATASET_FULL_PATH)
```
jsh2134, line 79: this is not obvious (and a bad implementation of unit testing on our part).....but TEST_DATASET_FULL_PATH is an actual real public dataset on SolveBio and we shouldnt modify it.

a cleaner test would be to create a new folder and test the metadata case with that.

folder = self.client.Object.get_or_create_by_full_path('~/test_folder_metadata', object_type='folder')

you can move this test into `test_objects.py`

solvebio/resource/apiresource.py
```diff
@@ -312,9 +312,21 @@ def save(self):
...
-        if obj._unsaved_values:
-            for k in obj._unsaved_values:
-                if k == 'id' or k == self.ID_ATTR:
-                    continue
-                params[k] = getattr(obj, k) or ""
+        unsaved_values = obj._unsaved_values
+
+        # adding 'metadata' attribute to be serialized
+        if getattr(obj, 'metadata'):
+            unsaved_values.add('metadata')
```
jsh2134, line 319: Maybe this is overkill,  but there are other fields that are nested and update-able. We could add a list to each `resource` class and loop through that list. Then it is not so "hard-coded"

`metadata` is the most common one, but a quick grep in the API code also shows `Beacon.params` and `SavedQuery.params` and lots of others (though they may not be editable once set)

solvebio/test/test_dataset.py
nikolamaric: This makes sense. Will do that.

solvebio/resource/apiresource.py
nikolamaric: I agree. That is the reason why I highlighted it was a "quick and dirty" solution. Let me think about this, but the idea with an additional list makes sense indeed.

jsh2134: ya on second though, we can add in the future if it comes up. 

Would be similar class attribute to the RESOURCE_VERSION or LIST_FIELDS https://github.com/solvebio/solvebio-python/blob/master/solvebio/resource/datasetmigration.py#L22

solvebio/test/test_object.py
```diff
@@ -236,3 +236,21 @@ def test_object_add_tag_non_iterable_or_string(self, ObjectCreate, ObjectSave):
         tags = 1
         file_.tag(tags)
         self.assertTrue(file_.has_tag(tags))
+
+    def test_object_set_metadata(self):
+        folder = self.client.Object.get_or_create_by_full_path('~/test_folder_metadata',
+                                                               object_type='folder')
```
jsh2134, line 242: since it is `get_or_create_` the metadata may already be set from a previous run of the test. can either create a new one each time...(use .create_by_full_path()) ...or one retrieving, clear the metadata dict and (do folder.save()) and then test that the set function works.

nikolamaric: Sure.

solvebio/test/test_dataset.py
nikolamaric: Done.

solvebio/resource/apiresource.py
nikolamaric: You are the boss... :)

PR #357, opened 2020-08-26
Join left explode

nikolamaric:


nikolamaric:
The tests fail, but I do not why? On my computer, they pass without issues.

jsh2134:
> The tests are failing, but I do not why? On my computer, they are passing without issues.

Tests run with a different user in an alternate test account, haven't investigated but that cold be the reason. Will send you information on how to run as that user.

solvebio/query.py
```diff
@@ -830,6 +834,12 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
         # If no key_b is provided, use the same key as for A
         key_b = key_b or key
 
+        # Limit cannot be larger than 1000 * 100
```
davecap, line 837: is a join limit needed?

solvebio/query.py
```diff
@@ -851,27 +861,41 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
...
+                        fields={},
+                        filters={},
+                        entities={},
+                        limit={}
+                    ) if get(record, "{}") else {{}}
+                """.format(query_b._dataset_id,
+                           [f.name for f in query_b_fields] + [key_b],
+                           filters,
+                           query_params.get('entities'),
+                           query_b._limit,
```
davecap, line 882: I don't know if this is really correct to use the query b limit for this purpose. We can set this to some large limit, like 100000.

solvebio/query.py
```diff
@@ -880,26 +904,38 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
...
                 "name": name,
                 "title": field.title,
                 "data_type": field.data_type,
                 "ordering": field.ordering,
-                "is_list": False,
+                "is_list": True,
                 "is_transient": False,
                 "expression": """
-                    get(record, "{}.{}")
+                    [get(item, "{}.{}") for item in record.exploded_join_field]
```
davecap, line 920: I'm not sure if the double explode is needed. Can't we just create a list of values using `[get(item, "...") for item in record. query_b_join_field_name`? and then explode in post_annotate

solvebio/query.py
```diff
@@ -880,26 +904,38 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
...
                 """.format(query_b_join_field_name, field.name),
-                "depends_on": [query_b_join_field_name]
+                "depends_on": [query_b_join_field_name, 'exploded_join_field']
             })
 
         # Add to any existing target fields
         new_query._target_fields += target_fields
 
-        return new_query
+        # Explode new_query records
```
davecap, line 928: the post_annotation_expression should be passed into the new_query object instead of annotating the output.

jsh2134:
Actually tests were failing because the new annotator feature had not been deployed to prod yet

solvebio/query.py
nikolamaric: Done.

solvebio/query.py
nikolamaric: You are right. Done.

solvebio/query.py
nikolamaric: I thought it could be useful, however, a user can always do slicing if he doesn't want to work with a total number of rows.
Removed.

solvebio/query.py
nikolamaric: This is necessary, let's assume that dataset B has 1m rows and query B is created against dataset B but only taking into account the first 1k rows. If we have query A with only one row consisting of the key which is in all 100 rows of query B, the result of the `join` (left outer join) method is 1 x 1k rows. However, without setting a limit to `query_b._limit` the result will be 1 x 100k rows, as `dataset_query` expression function returns up to 100k records, otherwise the exception will be raised as it is initially limited to that number.

So I have changed it a bit. If query B limit has been set up, that is used for limiting `dataset_query` otherwise limit is 100000.

solvebio/query.py
```diff
@@ -880,26 +884,31 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
...
+                """.format(field.name, query_b_join_field_name),
                 "depends_on": [query_b_join_field_name]
             })
 
         # Add to any existing target fields
         new_query._target_fields += target_fields
 
-        return new_query
+        # Explode new_query records
+        return new_query.annotate(
```
davecap, line 909: What this does is, it will submit the first query and then for each chunk of records, it will issue another set of API requests to annotate them.

Instead of these extra requests, what you can try is:

```
new_query._annotator_params = {"post_annotation_expression": "explode..."}
```

This will run the post_annotation_expression right in the same API request that the query is running.

davecap: Then you can just return new_query

solvebio/query.py
```diff
@@ -880,26 +884,31 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
...
                 "title": field.title,
                 "data_type": field.data_type,
                 "ordering": field.ordering,
-                "is_list": False,
+                "is_list": True,
                 "is_transient": False,
                 "expression": """
-                    get(record, "{}.{}")
-                """.format(query_b_join_field_name, field.name),
+                    [get(item, "{}") for item in record.{}]
```
davecap, line 900: might be safer to use `get(record, "{}")` for the query_b_join_field_name

solvebio/query.py
```diff
@@ -880,26 +884,31 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
             else:
                 name = field.name
 
+            query_b_prefixed_fields.append(name)
```
davecap, line 887: can you rename this to "explode_fields" ?

solvebio/query.py
nikolamaric: I tried that, but it wasn't give the right result... Let me check once again.

solvebio/query.py
davecap: But the dataset_query expression will only return 1k records in that case (since dataset B only has 1k), regardless of the limit. I don't see how it would return 100k.

solvebio/query.py
```diff
@@ -852,11 +852,11 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=True):
 
         # Prepare the filters for the B query expression
         query_params = query_b._build_query()
-        base_filter = '["{}", get(record, "{}")]'.format(key_b, key)
+        # base_filter = '["{}", get(record, "{}")]'.format(key_b, key)
```
davecap, line 855: I think this is the critical part of the join, as far as I remember. Why is it being removed?

davecap: Moving it into the expression down below makes the join very inefficient as it will pull all records and filter in the annotator as opposed to within Elasticsearch.

nikolamaric: if base_filter stays, `join()` does not behave as a true left join. For each record, it will query a new dataset according to filters, which is not what we expect to have when the `query_b` has been created.

solvebio/query.py
```diff
@@ -435,9 +435,8 @@ def __len__(self):
                  SELECT * FROM <table> [WHERE condition] [LIMIT number]
               )
         """
-        if getattr(self, '_total', None) == float('inf'):
-            # TODO: Is there a better way?
-            return 999999999
+        if getattr(self, '_is_join', False):
+            return len(self._buffer)
```
davecap, line 439: So the idea here is that the buffer is never empty if its a join, right?

nikolamaric: It does not matter, if query_a is empty the result is empty join. If query_b is empty the result is all rows from query_a but with empty values.

solvebio/query.py
```diff
@@ -627,13 +626,7 @@ def next(self):
...
         self._buffer_idx += 1
 
-        try:
-            return self._buffer[self._buffer_idx - 1]
-        except IndexError:
-            # If this is a join query, there might not be any more
-            # items in the buffer, and no more data to fetch.
-            # TODO: Should we confirm it is a join?
-            raise StopIteration()
+        return self._buffer[self._buffer_idx - 1]
```
davecap, line 629: Won't you still get an exception here?

nikolamaric: No exceptions, I added 2 tests in order to cover empty query_a/query_b. Seems it works well.

nikolamaric:
The PR is ready for review.

davecap:
A couple things that would be nice to fix or discuss:

1. If the key is the same for both datasets, maybe we can just omit the `b_(key)`, for example if the join is on `variant` and both datasets have `variant`, just skip the `b_variant` field?
2. I noticed in the example that join results without any records from the B query show up with empty lists:

```
{'count': [], 'cosmic_id': [], 'variant': 'GRCH37-10-89692776-89692778-A', 'b_variant': [], 'gene': 'PTEN'}
```

I wonder if its possible to avoid this?

nikolamaric:
> A couple things that would be nice to fix or discuss:
> 
> 1. If the key is the same for both datasets, maybe we can just omit the `b_(key)`, for example if the join is on `variant` and both datasets have `variant`, just skip the `b_variant` field?
> 2. I noticed in the example that join results without any records from the B query show up with empty lists:
> 
> ```
> {'count': [], 'cosmic_id': [], 'variant': 'GRCH37-10-89692776-89692778-A', 'b_variant': [], 'gene': 'PTEN'}
> ```
> 
> I wonder if its possible to avoid this?

1. it makes sense
2. this is a very good question, I have to take a look at this but as far as I remember there were some difficulties doing that

nikolamaric:
The PR can be closed as a branch from #363 was created from this branch and all commits are already there now.

Closed.

PR #363, opened 2020-09-21
Add ability to query files and join datasets

nikolamaric:


solvebio/resource/object.py
```diff
@@ -545,3 +549,19 @@ def untag(self, tags, dry_run=False, apply_save=True):
...
+    def query_object_content(self, limit=100, **_kwargs):
+
+        if not self.is_file:
+            raise SolveError('The functionality is only supported for files. '
+                             'This is a {}.'.format(self.object_type))
+
+        # TODO: limit param has been implemented in the API endpoint so far
+        params = {'limit': limit}
+
+        try:
```
davecap, line 562: Is it necessary to try/except this if there is just a raise?

solvebio/resource/object.py
```diff
@@ -545,3 +549,19 @@ def untag(self, tags, dry_run=False, apply_save=True):
         """Remove tags on an object"""
 
         return self.tag(tags=tags, remove=True, dry_run=dry_run, apply_save=apply_save)
+
+    def query_object_content(self, limit=100, **_kwargs):
```
davecap, line 553: please use `kwargs` instead of `_kwargs` to keep the code style consistent

solvebio/resource/object.py
nikolamaric: sure, this happens when you copy/paste things without thinking about them.

solvebio/resource/object.py
```diff
@@ -553,23 +553,19 @@ def untag(self, tags, dry_run=False, apply_save=True):
...
-
         if self.is_dataset:
-            params_.update(params)
-            return self.query(query=query, **params_)
+            return Dataset(self.dataset_id, client=self._client).query(**params)
         else:
-            if not self.size or not isinstance(self.size, int):
-                raise SolveError('An empty file {} cannot be queried.'.format(self.filename))
-
             # TODO: limit param has been implemented in the S3 Select API endpoint so far
```
davecap, line 567: can remove this comment

nikolamaric: done

nikolamaric:
The PR is basically ready for review. The only thing which is left is documentation, but I can do it when the review has been done.

Note: the PR depends on #357 as it was created from that branch.

solvebio/query.py
```diff
@@ -941,3 +949,300 @@ def execute(self, **params):
         _params.update(**params)
         response = self._client.post('/v2/batch_query', _params)
         return response
+
+
+class QueryFile(object):
```
davecap, line 954: In hindsight, it looks like it would be simple to create a base class (QueryBase) as a mixin class with everything in QueryFile (except __init__, _clone, and _build_query). Everything looks exactly the same.

How about creating a `class QueryBase(object)` and inheriting that from `Query` and `QueryFile`?

solvebio/resource/object.py
```diff
@@ -285,6 +285,9 @@ def upload_file(cls, local_path, remote_path, vault_full_path, **kwargs):
         _client = kwargs.pop('client', None) or cls._client or client
 
         local_path = os.path.expanduser(local_path)
+        # TODO: the if statement below can be deleted when
```
davecap, line 288: Seems like we can remove this comment

solvebio/resource/object.py
```diff
@@ -294,7 +297,8 @@ def upload_file(cls, local_path, remote_path, vault_full_path, **kwargs):
 
         # Get MD5, mimetype, and file size for the object
         local_md5, _ = md5sum(local_path, multipart_threshold=None)
-        _, mimetype = mimetypes.guess_type(local_path)
+        mime_tuple = mimetypes.guess_type(local_path)
+        mimetype = mime_tuple[1] if mime_tuple[1] else mime_tuple[0]
```
davecap, line 301: please add a comment that explains what is happening here

solvebio/resource/object.py
```diff
@@ -545,3 +553,20 @@ def untag(self, tags, dry_run=False, apply_save=True):
...
+
+    def query(self, **params):
+        """
+        Return the Query or QueryFile object depending on object type
+        that represents query results against an object.
+        """
+        from solvebio.resource.dataset import Dataset
+        from solvebio.query import QueryFile
+
+        if not self.is_file and not self.is_dataset:
```
davecap, line 565: Personal preference, but how about:

```
if self.is_dataset:
   ...
elif self.is_file:
   ...
else:
   raise SolveError(...)
```

davecap:
Overall looks good. Just wondering it can some how be separated from the join PR changes? It looks like some of the changes from there are in here. Or we can merge this into the join PR if thats easier.

solvebio/query.py
nikolamaric: If you can recall, I suggested something similar to that, with a difference that QueryBase should have been an abstract class. However, when I think again it is not necessary to be the abstract class.

davecap: Yea I remember. I just looked through and it seems that most of the methods are exactly the same, so you can just use a mixin class.

nikolamaric:
The PR was created from #357 so when this PR is closed the parent PR can be closed as well.

Closes #357

jsh2134 (approved):

davecap (approved):

PR #364, opened 2020-09-24
Add solvebio-recipes tool

damnjanovictanja:
Add: Testing of the sync_recipes script through pytest, instead of nose (python path needed to include current dir for this)
Recipes tests only for python3, couldn't make fixtures work with python2
Ref: Different representation for unicode in RecipeDumper for python 2, otherwise 'item': !!python/unicode "some string" is dumped
Add: sync_recipes as a new entry point to setup.py
Usage example: solvebio-recipes --mode "sync" --all public_recipes.yml


setup.py
```diff
@@ -6,7 +6,7 @@
 import warnings
 
 VERSION = 'undefined'
-install_requires = ['six', 'pyprind']
+install_requires = ['six', 'pyprind', 'pyparsing==2.4.7']
```
jsh2134, line 9: what uses pyparsing? is this still needed?

recipes/public.yml
```diff
@@ -0,0 +1,75 @@
...
+    version: 1.0.0
+    is_public: true
+    fields:
+      name: gene
+      data_type: string
+      entity_type: gene
+      is_list: true
+      expression: |
+        dataset_field_values(
+                'solvebio:public:/dbSNP/2.0.0-b153/Variants-{}'.format(get(record, 'genomic_coordinates.build')),
```
jsh2134, line 72: can you change to `2.0.0-b151` for now until `2.0.0-b153` is ready

damnjanovictanja: Changed to b151

setup.py
damnjanovictanja: Not used, removed as a requirement

PR #367, opened 2020-10-07
Sync recipes command refactoring

damnjanovictanja:
Sync, delete and export modes can be used as subcommands:
solvebio-recipes sync --all path/to/yml
solvebio-recipes delete --name "dbsnp (v1.0.0)"
solvebio export --account-recipes path/to/yml


jsh2134 (approved):
Looks great. I'll give it a go and test it out

PR #371, opened 2020-10-09
Fix for multiple join errors

nikolamaric:
- Fix the prefix issue when two fields after adding a prefix have the same name
- Add newly prefixed fields to the `explode_fields` list in order to be taken into an account for exploding.

### DIFF
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -854,8 +854,7 @@ class Query(QueryBase):
         new_query = self._clone()
 
         # Initialize _explode_fields attribute if it does not exist
-        new_query._explode_fields = [] if not getattr(self, '_explode_fields', None) \
-            else self._explode_fields
+        new_query._explode_fields = getattr(self, '_explode_fields', None) or []
 
         # Set list of existing field names to avoid overwriting fields
         # in the join.
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -852,6 +852,10 @@ class Query(QueryBase):
 
         # Prepare new returned query object
         new_query = self._clone()
+
+        # Set prefix in order to use it in a mutiple join
+        new_query._prefix = join_id + '_'
+
         # Set list of existing field names to avoid overwriting fields
         # in the join.
         existing_field_names = [f.name for f in self.fields()]
@@ -914,7 +918,10 @@ class Query(QueryBase):
             explode_fields.append(name)
 
             if name in existing_field_names:
-                logger.warning("Field '{}' will be overwritten in the join results".format(name))
+                _name = new_query._prefix + field.name
+                logger.warning("Field '{}' will be overwritten in the join results as '{}'. ".
+                               format(name, _name))
+                name = _name
 
             target_fields.append({
                 "name": name,
@@ -932,6 +939,9 @@ class Query(QueryBase):
 
         # Add to any existing target fields
         new_query._target_fields += target_fields
+
+        # Extend explode_fields with prefixed fields to be exploded
+        explode_fields = [field['name'] for field in new_query._target_fields if 'join' not in field['name']]
 
         new_query._annotator_params = {
             'post_annotation_expression': "explode(record, fields={})".format(explode_fields)
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -941,8 +941,8 @@ class Query(QueryBase):
         new_query._target_fields += target_fields
 
         new_query._annotator_params = {
-            'post_annotation_expression': "explode(record, fields={})".
-                format(new_query._explode_fields)
+            'post_annotation_expression':
+                "explode(record, fields={})".format(new_query._explode_fields)
         }
         new_query._is_join = True
         return new_query
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -874,7 +874,7 @@ class Query(QueryBase):
 
         # If a joining key in both datasets is the same and
         # it is not the only one field then remove it from query_b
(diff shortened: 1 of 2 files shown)


davecap:
should this branch be based off the other branch?

nikolamaric:
Damn!

davecap:
You can change the base, click "edit" at the top

davecap:
Hmm it looks like there are commits from master that are not yet in the base branch, can you make sure the branches are both up to date with master?

nikolamaric:
I think now it is good.

solvebio/query.py
```diff
@@ -852,6 +852,10 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=False):
 
         # Prepare new returned query object
         new_query = self._clone()
+
+        # Set prefix in order to use it in a mutiple join
+        new_query._prefix = join_id + '_'
```
davecap, line 857: whats the purpose of setting it to the _prefix attribute if it is only used in this function?

solvebio/query.py
```diff
@@ -914,7 +918,10 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=False):
             explode_fields.append(name)
 
             if name in existing_field_names:
-                logger.warning("Field '{}' will be overwritten in the join results".format(name))
```
davecap, line 917: How about this, instead of just showing a warning, lets raise an exception that says `Field '{}' found in both queries, please use a different prefix` then the user is just forced to set a new prefix and the query wont crash.

solvebio/query.py
```diff
@@ -871,8 +872,9 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=False):
         # Try to use a unique field for the sub-query data
         query_b_fields = query_b.fields()
 
-        # If a joining key in both datasets is the same then remove it from query_b
-        if key_b == key:
+        # If a joining key in both datasets is the same and
+        # it is not the only one field then remove it from query_b
+        if key_b == key and len(query_b_fields) == 1 and query_b_fields[0].name != key_b:
```
davecap, line 877: What issue does this fix?

solvebio/query.py
```diff
@@ -933,6 +936,9 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=False):
         # Add to any existing target fields
         new_query._target_fields += target_fields
 
+        # Extend explode_fields with prefixed fields to be exploded
+        explode_fields = [field['name'] for field in new_query._target_fields if 'join' not in field['name']]
```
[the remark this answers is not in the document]
nikolamaric: without that, the result will contain lists as values for all previous joins before the last one. 

```
{'c_id': '1', 'name': 'Francis Crick', 'b_id': ['1'], 'b__commit': ['1329009293354399140'], 'c__commit': '1329009293354399140', 'b_name': ['Francis Crick'], '_commit': '1329011331274567012', 'id': '1', 'c_name': 'Francis Crick'}
{'c_id': '2', 'name': 'James Watson', 'b_id': ['2', '2'], 'b__commit': ['1329009293354399140', '1329009293354399140'], 'c__commit': '1329009293354399140', 'b_name': ['James Watson', 'James Watson'], '_commit': '1329011331274567012', 'id': '2', 'c_name': 'James Watson'}
{'c_id': '2', 'name': 'James Watson', 'b_id': ['2', '2'], 'b__commit': ['1329009293354399140', '1329009293354399140'], 'c__commit': '1329009293354399140', 'b_name': ['James Watson', 'James Watson'], '_commit': '1329011331274567012', 'id': '2', 'c_name': 'James Watson'}
```

`explode` function requires fields that will be parsed in order to make tuples with their values. 

https://github.com/solvebio/api.solvebio.com/blob/c375407489cd4c0da7c4fc561ed016138fa6036a/api/annotations/functions/records.py#L21

solvebio/query.py
nikolamaric: e.g. https://github.com/solvebio/solvebio-python/blob/bf3394638d5ce39a52e97d6284e893c4744dbfd8/solvebio/test/test_query.py#L378

This will not work if we remove the `gene` from query_b because it is the only field in that query.

nikolamaric: I updated the link in a previous comment.

solvebio/query.py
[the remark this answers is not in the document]
davecap: Ah ok, need to think about this some more then! Maybe we should store the list of explode fields in the query and pass that through to the chained queries? would that work? i.e. if `self._join_fields` then add those to the explode fields list. I don't like this check for 'join' in the field name because it will break if the field has join in it.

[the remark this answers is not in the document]
nikolamaric: Seems it works. I added them to `self._explode_fields` attribute that is going to be extended by all fields that are created within each additional join function.

solvebio/query.py
```diff
@@ -853,6 +853,10 @@ def join(self, query_b, key, key_b=None, prefix="b_", always_prefix=False):
         # Prepare new returned query object
         new_query = self._clone()
 
+        # Initialize _explode_fields attribute if it does not exist
+        new_query._explode_fields = [] if not getattr(self, '_explode_fields', None) \
```
davecap, line 857: How about:

```
new_query._explode_fields = getattr(self, '_explode_fields', None) or []
```

nikolamaric: Yes, much elegant solution :)

davecap (approved):

davecap:
Works great as far as I can tell :)

PR #373, opened 2020-10-20
Scoped object search

jsh2134:
Search from within the folders/files/dataset shortcuts in the same way as `Object.search`, but scope it to the shortcut in both location and object type

The following searches can be replaced with one nice command

```
folder.search("foobar AND type:file", parent_object_id=folder.id)
folder.search("foobar", object_type="file", parent_object_id=folder.id)
folder.files(query="foobar", parent_object_id=folder.id)
```

become
```
folder.files("foobar")
```

### DIFF
--- a/solvebio/resource/object.py
+++ b/solvebio/resource/object.py
@@ -421,6 +421,11 @@ class Object(CreateableAPIResource,
     def ls(self, **params):
         return self.objects(**params)
 
+    def search(self, query='', **params):
+        params['parent_object_id'] = self.parent_object_id
+        params['object_type'] = self.object_type
+        return super(Object, self).search(query, **params)
+
     def __getattr__(self, name):
         """Shortcut to access attributes of the underlying Dataset resource"""
         from solvebio.resource import Dataset
--- a/solvebio/resource/object.py
+++ b/solvebio/resource/object.py
@@ -388,7 +388,7 @@ class Object(CreateableAPIResource,
 
         return obj
 
-    def _object_list_helper(self, **params):
+    def _object_list_helper(self, query='', **params):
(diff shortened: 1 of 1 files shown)


solvebio/resource/object.py
```diff
@@ -406,20 +406,39 @@ def _object_list_helper(self, **params):
...
+    def ls(self, query='', **params):
+        query = query or params.pop('query', '')
+        return self.objects(query=query, **params)
+
+    def search(self, query='', **params):
+        query = query or params.pop('query', '')
+        # TODO
+        params['ancestor_id'] = self.id
+        params['limit'] = 1000
+        return super(Object, self).search(query, **params)
```
jsh2134, line 441: I believe I left this TODO in here in order to determine if there should be a recursive option. Should the user be able to get anything in this folder, or just depth=0 objects.

 and/or `ancestor_id` needs to be replaced by `parent_id` (although the server may handle this)

This PR needs tests

solvebio/resource/object.py
```diff
@@ -406,20 +406,39 @@ def _object_list_helper(self, **params):
...
         return self._object_list_helper(**params)
 
-    def ls(self, **params):
-        return self.objects(**params)
+    def ls(self, query='', **params):
+        query = query or params.pop('query', '')
+        return self.objects(query=query, **params)
+
+    def search(self, query='', **params):
+        query = query or params.pop('query', '')
```
davecap, line 437: This is not necessary

solvebio/resource/object.py
davecap, line 441: Object.search(), super not necessary because search is a classmethod

PR #377, opened 2020-11-10
Add support for archiving and restoring datasets

jsh2134:
- Add DatasetSnapshot and DatasetRestore tasks
- Add dataset.archive() and dataset.restore() shortcuts


solvebio/resource/object.py
```diff
@@ -566,3 +566,36 @@ def query(self, **params):
...
+        """
+        Restore this dataset
+        """
+        if not self.is_dataset:
+            raise SolveError("Only datasets can be restored.")
+
+        # Updating storage class is only available at the objects endpoint
+        # TODO this doesnt work if it is None
+        self.storage_class = storage_class
+        self.save()
```
jsh2134, line 598: restore to "Standard" if no storage class passed. add comments to get last known storage class or vault default or account default

solvebio/resource/dataset_snapshot_task.py
```diff
@@ -0,0 +1,54 @@
...
+        ('status', 'Status'),
+        ('created_at', 'Created'),
+    )
+
+    @property
+    def dataset(self):
+        response = self._client.get(self['dataset']['url'], {})
+        return convert_to_solve_object(response, client=self._client)
+
+    def follow(self, loop=True, sleep_seconds=Task.SLEEP_WAIT_DEFAULT):
```
davecap, line 28: Is there some way to abstract these follow() methods? maybe with some kind of ActivityResource or FollowableResource or something?

jsh2134: yes, it could be added. each Task would need to define a custom "progress" function though. Also some tasks (imports/migrations) have commits that need to be followed

davecap: Not a big deal, can always refactor later.

solvebio/resource/dataset_restore_task.py
```diff
@@ -0,0 +1,54 @@
...
+                    print("Restore '{0}' is {1}".format(self.id, self.status))
+
+            if not loop:
+                return
+
+            time.sleep(sleep_seconds)
+            self.refresh()
+
+        if self.status == 'completed':
+            print('Restore complete! Dataset is now avaiable')
```
davecap, line 54: Typo. Maybe we don't need these prints?

davecap: Or we can keep them

davecap (approved):

solvebio/resource/dataset_restore_task.py
nikolamaric: I think we should keep them, one small typo `print('Restore complete! Dataset is now avaiable')`  -> `available`

nikolamaric (approved):

PR #382, opened 2020-12-07
Enable using filters, fields, exclude_fields for a QueryFile object.

nikolamaric:


nikolamaric:
closes #383

solvebio/query.py
```diff
@@ -1093,11 +1189,24 @@ def execute(self, offset=0, **query):
 
         # If the request results in a SolveError (ie bad filter) set the error.
         try:
-            self._response = self._client.get(self._data_url, _params)
+            # TODO: Think if it makes sense to use teh GET S3 Select endpoint anymore
```
davecap, line 1192: Agree, we can remove from here

davecap (approved):
Lets wait to get the tests passing before merging this

solvebio/query.py
```diff
@@ -1088,19 +1132,31 @@ def execute(self, offset=0, **query):
...
+            self._response = self._client.post(self._data_url, _params)
         except SolveError as e:
             self._error = e
             raise
 
-        logger.debug('query response took: %(took)d ms, total: %(total)d'
-                     % self._response)
+        logger.debug('query response took: {} ms, total: {}'.
+                     format(self._response['took'],
+                            'nan' if self._response['total'] is None else self._response['total']))
```
davecap, line 1150: It is probably ok to just print the contents of [total] as-is, since it is a debug log.

nikolamaric: Done.

solvebio/query.py
```diff
@@ -1009,33 +1009,53 @@ def execute(self, **params):
...
+    DEFAULT_PAGE_SIZE = 1000
 
     def __init__(
             self,
             file_id,
-            limit=QueryBase.DEFAULT_PAGE_SIZE,
+            fields=None,
+            exclude_fields=None,
+            filters=None,
+            limit=DEFAULT_PAGE_SIZE,
```
davecap, line 1024: set limit to inf

solvebio/query.py
```diff
@@ -264,7 +264,8 @@ def __len__(self):
                  SELECT * FROM <table> [WHERE condition] [LIMIT number]
               )
         """
-        return min(self._limit, self.count()) if not self.count() is None else self._limit
+        _limit = 10 ** 15 if self._limit == float('inf') else self._limit
```
davecap, line 267: I guess the issue here is that because we don't have `total`, it tries to send `inf` to the API and fails somewhere? So we need to set a large value for limit here. Are there any other options instead of setting to 10**15?

davecap: What happens if we use  `None`?

nikolamaric: for example `__len__` is called here: https://github.com/solvebio/solvebio-python/blob/42e8aab4334c9bc71a88744b7f67db8bc3ea198e/solvebio/query.py#L413

and `None` raises an error `TypeError: 'NoneType' object cannot be interpreted as an integer`.

Honestly, I do not have an idea of what we can add there instead of some large value?

davecap: What if we handle both cases where total and _limit are both None in the iterator, instead of wrapping it up in `len(self)`?

nikolamaric: It means that we have to do the same wherever `len(self)` is currently called in code, right?

davecap: What if in __len__ we `return None if _limit is None and self.count() is None`

nikolamaric: it is possible, but as far as I know instead of calling `len(self)` we have to call it like `self.__len__()` otherwise it will raise `TypeError: 'NoneType' object cannot be interpreted as an integer`

davecap: Ok can you set this magic value as a constant in the function `INF = 10**15` and add a comment above it to explain

davecap: Actually, last thing. Can we just set the initial value of limit to INF instead of float(inf)?

davecap:
Great, will merge when tests pass!

jsh2134:
😻

PR #386, opened 2020-12-28
Handle a query slicing when the potential number of returned records in the query is larger than the number of records left in the dataset/file

### LINKED ISSUE #385: Dataset/File query raises an error when it has been sliced and a stop param is set to float('inf')
Slicing results of dataset/file query raises an error when `stop` param is set to `float('inf')`, or in other words to the end of the result list.

```
Example:

d = Dataset.retrieve(1390529638872624453)
for i in d.query()[1:]:
    pprint(i)

Result:

{'_commit': '1390529745255123200',
 '_id': 'AXX6NkquZs-hDkFUZ3lY',
 'city': 'Chicago',
 'name': 'Kate',
 'occupation': 'Developer',
 'phonenumber': '(949) 555-6707'}
Traceback (most recent call last):
  File "C:/NikolaM/Projects/SolveBio/Repos/solvebio-python/test_examples/test_archiving_restore.py", line 8, in <module>
    for i in d.query(page_size=1)[1:]:
  File "C:\NikolaM\Projects\SolveBio\Repos\solvebio-python\solvebio\query.py", line 392, in __next__
    return self.next()
  File "C:\NikolaM\Projects\SolveBio\Repos\solvebio-python\solvebio\query.py", line 422, in next
    return self._buffer[self._buffer_idx - 1]
IndexError: list index out of range
```

nikolamaric:
Slicing results of dataset/file query raises an error when the `stop` param is set to `float('inf').

closes #385

### DIFF
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -409,7 +409,12 @@ class QueryBase(object):
             self.__iter__()
 
         # len(self) returns `min(limit, total)` results
-        if self._cursor == len(self):
+        # Raise StopIteration when query()[n:] has been exhausted
+        if self._slice and self._slice.stop == float('inf') and self._cursor == len(self) - 1:
+            raise StopIteration()
+
+        # Raise StopIteration when query()[n:m] has been exhausted
+        elif self._cursor == len(self):
             raise StopIteration()
 
         if self._buffer_idx == len(self._buffer):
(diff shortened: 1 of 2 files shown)


jsh2134 (approved):
LGTM. Can you add a quick regression test for this (the one you showed on the call) ?

nikolamaric:
Sure.

jsh2134 (approved):

PR #387, opened 2021-01-05
Handle parallel migrations

davidhpark:
* Updates follow logic to accommodate parallel migrations


solvebio/query.py
```diff
@@ -827,7 +827,7 @@ def migrate(self, target, follow=True, **kwargs):
             **kwargs)
 
         if follow:
-            migration.follow()
```
davidhpark, line 830: @jsh2134 like this??

davecap (approved):

jsh2134 (approved):

PR #388, opened 2021-01-14
Improve condition that checks page_size param

nikolamaric:
Since `page_size` param validation only checks if it is less than 0, in case it is set to a value greater than 10 000, the error message is a bit unclear which is the result of the internal logic that handles `query` method params which are sent to the API. 
https://github.com/solvebio/solvebio-python/blob/e154d4112fe1d52033f54def4d7111c970d53445/solvebio/query.py#L722

e.g.

```
ds = Dataset.retrieve(1052228458696994189)
q = ds.query(page_size=20000)

# response
solvebio.errors.SolveError: Bad Request (https://solvebio.api.solvebio.com/v2/datasets/1052228458696994189/data)
Error: {'limit': ["'limit' must be an integer between 0 and 10000"]})
```

The error message says that `limit` has been exceeded even though it wasn't set in the `query` method, only `page_size`.

This issue is applied to both dataset and file query as they utilize the same internal logic.


davecap:
I'm hesitant to merge this since it couples the client to a parameter in the API. I feel that it is an advanced usage pattern and so we can just document the issue if needed.

davecap (approved):

PR #389, opened 2021-01-19
Enable 'output_format' for a file-based queries 

nikolamaric:
Set `output_format` to `json` by default and enable its setting in the `query` method.


davecap:
Has this been tested with any format other than json? I don't think it works with csv/tsv.

nikolamaric:
> Has this been tested with any format other than json? I don't think it works with csv/tsv.

It has been tested, and worked like a charm (I have done it now)

nikolamaric:
fixes #392

nikolamaric:
Seems we have some issues with Travis CI, as it has been stuck since Friday?

davecap:
Ah yes, sorry travis has actually been replaced for the GH Actions. We can just ignore it.

nikolamaric:
The PR is merged into #394. Closing.

PR #390, opened 2021-01-28
Migrate from Travis to GitHub actions

davecap:


davecap:
I think we may not need both push and pull_request triggers. But I'm not exactly sure..

jsh2134:
wow GH Actions is so fast!

damnjanovictanja (approved):

jsh2134 (approved):

PR #391, opened 2021-02-08
Enable sorting of the object fields

nikolamaric:


nikolamaric:
Closing, we do not need this.

PR #394, opened 2021-03-02
Enable pagination for the join method

nikolamaric:


nikolamaric:
I merged this guy #389 into the PR as I needed logic for the `print` message here in order to cover all possible cases, therefore the merged PR won't be needed anymore when we merge this one.

nikolamaric:
Unfortunately, slicing doesn't work for the join function, as discussed we are going to add a note in docs about that.

solvebio/query.py
```diff
@@ -413,11 +432,16 @@ def next(self):
             self.__iter__()
 
         # len(self) returns `min(limit, total)` results
-        if self._cursor == len(self):
+        if not getattr(self, '_is_join', False) and self._cursor == len(self):
```
davecap, line 435: Can these be changed to remove the getattr now?

solvebio/query.py
```diff
@@ -1030,6 +1064,8 @@ def __init__(
             result_class=dict,
             debug=False,
             error=None,
```
davecap, line 1066: can you remove this? I don't think it is used actually (for input purposes)

Also, reorder the docstring so it matches the order of the params.

davecap (commented):
Looks good so far

davecap:
Feel free to close #389 then

solvebio/query.py
nikolamaric: I think it should stay as is because it is relevant for the regular dataset and file-based queries. 
It is not used for the join query as using `_cursor` does not make sense in that case since `len(self)` for join queries does not return lenght of `total` but `_buffer`. 
`next()` method is a common iterator method for all types of queries.

davecap: Ok! sounds good

solvebio/query.py
nikolamaric: Done.

solvebio/query.py
```diff
@@ -412,12 +431,20 @@ def next(self):
...
+
         # len(self) returns `min(limit, total)` results
-        if self._cursor == len(self):
+        if not _is_join and self._cursor == len(self):
             raise StopIteration
 
         if self._buffer_idx == len(self._buffer):
-            self.execute(self._page_offset + self._buffer_idx)
+            if _is_join:
+                if self._next_offset >= self._limit:
```
davecap, line 443: Add a comment:

# Since joins can return more results than we expect (due to `explode`)
# manually ensure that we haven't gone above the requested limit (default inf)

davecap (approved):

solvebio/query.py
```diff
@@ -292,10 +292,29 @@ def __repr__(self):
...
-                     aligns=['right', 'left'], sort=True),
-            pretty_int(len(self) - 1))
+        placeholder = 'many more results (total unknown)'
+
+        if getattr(self, '_is_join', False):
+            return '\n%s\n\n... %s.' % (
+                tabulate(list(self._buffer[0].items()), ['Fields', 'Data'],
+                         aligns=['right', 'left'], sort=True),
+                placeholder)
+        elif not getattr(self, '_output_format', False):
```
davecap, line 302: I think this may be wrong now since _output_format defaults to json

davecap (changes requested):

solvebio/query.py
nikolamaric: All is good, however, we can use `hasattr` instead of `getattr`  because we do not need value.

davecap: Ok, I did notice that file-based queries on JSON files now print the rows and not the JSON objects

```
f = Object.retrieve("XYZ")
f.query()

|          Fields | Data                                                                                            |
|-----------------+-------------------------------------------------------------------------------------------------|
|   configuration | 1                                                                                            |
|           debug | False                                                                                           |
| output_dir_path | xyz                                                                               |
|          sample | {...} |                                                             |
|         version | 3
```

Now gives:

```
|   Row | Data                                                                                                      |
|-------+-----------------------------------------------------------------------------------------------------------|
|     0 | {u'sample': {... |

```

nikolamaric: Ohh, yes! I changed that to be in accordance with the new file preview in the UI. Please look at this[PR](https://github.com/solvebio/www.solvebio.com/pull/1953). 

It can be changed easily if you want.

nikolamaric: image

davecap: Ok, I see that. I feel like it should be similar to how dataset.query() works, at least for JSON files. Doesn't really matter that much though. I'm ok with how it is now.

davecap (approved):

PR #405, opened 2021-11-03
remove dotenv from the streamlit lib

davecap:


krivi95 (approved):

solvebio/contrib/streamlit/app_example.py
```diff
@@ -1,7 +1,13 @@
 import streamlit as st
 import pandas as pd
 
+from dotenv import load_dotenv
+
+# Load environment variables from .env file
+load_dotenv()
```
krivi95, line 7: Maybe to remove the dotenv from requirements as well?
https://github.com/solvebio/solvebio-python/blob/master/solvebio/contrib/streamlit/dev-requirements.txt#L4

davecap: Yea we can, I figured it in the dev requirements so not a big deal either way.

PR #410, opened 2021-12-10
Add Global Search and Global Beacon support

### LINKED ISSUE #409: Add Global Beacon support
The new (beta) Global Beacon system currently has three main functions:

- Enable
- Disable
- Status

In the future there might be more direct querying options but for now we just need to implement the above.

davecap:
closes #409

## Description
Adding support:

- Global Beacon functionality:
  - Toggle on/off global beacon for the dataset
  - Check Global Beacon indexing status
- Global Search functionality
  - Low level search functionality, similar as there is on the Mesh web application ([Global Search docs](https://docs.solvebio.com/search/))
- Example Jupiter notebooks files for demonstrating new functionality that is introduced:
  - [global_beacon_idexing.ipynb](https://github.com/solvebio/solvebio-python/blob/issue-409-global-beacon-basics/examples/global_beacon_indexing.ipynb)
  - [global_search.ipynb](https://github.com/solvebio/solvebio-python/blob/issue-409-global-beacon-basics/examples/global_search.ipynb)

## Examples

- Global Beacon
  ```python
  # Dataset object
  dataset = solvebio.Object.retrieve(123456789)
 
  # Global Beacon functions
  dataset.enable_global_beacon()
  dataset.disable_global_beacon()
  dataset.get_global_beacon_status()
  ```
- Global Seach
  ```python
  # Advanced search
  advanced_query_results = GlobalSearch("fuji")
 
  # Entity search example
  GlobalSearch(entities=[["variant", "GRCH38-7-140753336-140753336-T"]])

  # Get list of subjects for the entity search
  GlobalSearch(entities=[["gene","BRCA2"]]).subjects()

  # Date range
  search = GlobalSearch()
  objects = search.filter(created_at__range=["2021-11-28","2021-12-28"])

  # Search for all datasets that are creted by the user Nikola
  f = Filter(type="dataset") & Filter(user="Nikola")
  results = GlobalSearch(filters=f)
  ```
  ```python
  # Chaining function calls for filter, entity, subjects
  s = GlobalSearch()

  for result in s.entity(gene="BRCA2"):
      print(result['id'])

  # Get subjects with BRCA2 in public vault
  s.filter(vault="public").entity(gene="BRCA2").subjects()

  # Get subjects count with BRCA2 in public vault
  s.filter(vault="public").entity(gene="BRCA2").subjects_count()

  # Get all vaults with BRCA2 datasets
  s.entity(gene="BRCA2").facets("vault")
  ```


examples/global_beacon_idexing.ipynb
```diff
@@ -0,0 +1,321 @@
+{
```
davecap, line 1: typo in this filename

examples/global_beacon_idexing.ipynb
```diff
@@ -0,0 +1,321 @@
...
+       "|              class_name | Object                                             |\n",
+       "|              created_at | 2021-11-29T11:24:42.093Z                           |\n",
+       "|     dataset_description | test                                               |\n",
+       "| dataset_documents_count | 26312                                              |\n",
+       "|       dataset_full_name | 1658666726768179211                                |\n",
+       "|              dataset_id | 1658666726768179211                                |\n",
+       "|                   depth | 1                                                  |\n",
+       "|             description | test                                               |\n",
+       "|         documents_count | 26312                                              |\n",
+       "|                filename | BEACONS-TEST-Genes-94-human-mouse-wOXjFS-1426764376|\n",
```
davecap, line 94: It might be nice to rename the dataset to something like `beacon-test-dataset`

davecap: also put it into a shared vault, I believe you cannot enable beacons on private vault datasets

examples/global_beacon_idexing.ipynb
```diff
@@ -0,0 +1,321 @@
...
+      ]
+     },
+     "execution_count": 3,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Getting the dataset\n",
+    "dataset = Object.retrieve(1658666726768179211)\n",
```
davecap, line 131: I would use get_by_full_path() in docs, it is more user friendly

examples/global_beacon_idexing.ipynb
```diff
@@ -0,0 +1,321 @@
...
+    "dataset.get_global_beacon_status()"
+   ]
+  }
+ ],
+ "metadata": {
+  "interpreter": {
+   "hash": "5365f12f12443ba93290371f268fae1547fcad5fb86506489681c9d374aefa71"
+  },
+  "kernelspec": {
+   "display_name": "Python 2.7.18 64-bit ('venv')",
```
davecap, line 301: Can you use python 3? not a big deal either way

solvebio/global_search.py
```diff
@@ -0,0 +1,159 @@
...
+          - `query` (optional): An optional query string (advanced search).
+          - `filters` (optional): Filter or List of filter objects.
+          - `entities` (optional): List of entity tuples to filter on (entity type, entity).
+          - `ordering` (optional): List of fields to order the results by.
+          - `limit` (optional): Maximum number of query results to return.
+          - `page_size` (optional): Number of results to fetch per query page.
+          - `result_class` (optional): Class of object returned by query.
+          - `debug` (optional): Sends debug information to the API.
+        """
+        self._data_url = '/v2/search'
```
davecap, line 43: Call super __init__ and then overwrite the data_url that way

solvebio/global_search.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if not 0 < self._page_size <= self.MAX_PAGE_SIZE:
+            raise Exception('\'page_size\' parameter must be in '
+                            'range [1, {}]'.format(self.MAX_PAGE_SIZE))
+
+        # Set up the SolveClient
+        # (kwargs overrides pre-set, which overrides global)
+        self._client = kwargs.get('client') or self._client or client
+
+    def _clone(self, filters=None, limit=None):
```
davecap, line 86: Can we just use the parent class methods?

solvebio/global_search.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # Add or modify query parameters
+        # (used by BatchQuery and facets)
+        q.update(**kwargs)
+
+        return q
+
+    def subjects(self):
+        """If entity seaarch is performed returns the list of subjects"""
+        if not self._entities:
```
davecap, line 152: I believe with the new code it might be possible to get subjects on any query.. but I could be wrong about this. Might still want to remove this short circuit

solvebio/global_search.py
```diff
@@ -0,0 +1,158 @@
+# -*- coding: utf-8 -*-
```
davecap, line 1: A few thoughts on this:

```
from solvebio import GlobalSearch

s = GlobalSearch()

# Add a new method for entity search similar to how filter() works?
for result in s.entity(gene="BRCA2"):
    print(result['id'])

# Get subjects with BRCA2 in some_vault
s.filter(vault="some_vault").entity(gene="BRCA2").subjects()

# Get subjects count with BRCA2 in some_vault
s.filter(vault="some_vault").entity(gene="BRCA2").subjects_count()
```

Is it possible to cast the response objects as `Object`?
What you would do is call `Object.construct_from(i)` for each logical object result (and Vault.construct_from(i) if the result is a vault) in the _response when it is set by `execute()`.

examples/global_beacon_idexing.ipynb
krivi95: I guess when you say shared vault you mean the `public` vault?

It seems that I don't have proper permissions on the `stag` environment to move the dataset from personal to public vault:
image

krivi95: @nikolamaric gave me permissions, works now.

examples/global_beacon_idexing.ipynb
krivi95: I had issues running `setup.py` with `python 3` in order to set up environment for development. With `python 2.7` it worked, so that's the reason why I settled down on using python 2.7 for the example notebooks as well.

Anyway, I've rerun the Jupyter Notebooks using python 3, everything works as expected.

solvebio/global_search.py
krivi95: As the parent class of `GlobalSearch` is `Query`, which has some additional attributes like required dataset id and optional parameters in constructors such as _genome_build, _annotator_params, _target_fields and others.

And the `_clone` method creates a new instance of that class, in our case a new instance of `GlobalSearch` (`self.__class__(params...)`). So because of that, we have to override that method and specify the correct arguments for the constructor for GlobalSearch class.

solvebio/global_search.py
krivi95: I know that `GlobalSearch` and `Query` classes are not quite the same, but they share a lot of common functionalities. I still believe that GlobalSearch should be derived from Query class, so we don't unnecessarily duplicate the code that's already in there.

davecap: Yes I agree, just trying to further remove some duplicate code. That approach above sounds fine to me

solvebio/global_search.py
davecap: Ah yea I see. Ok thats fine.

solvebio/global_search.py
```diff
@@ -148,10 +151,36 @@ def _build_query(self, **kwargs):
...
+        existing set with AND.
+
+        kwargs can contain only one entity, entity_type as parameter name and entity as its value.
+        If entity is already set for the GlobalSearch, it will be overridden.
+        """
+
+        if not kwargs:
+            raise AttributeError('Faceting requires at least one field')
+
+        if len(kwargs) > 1:
```
davecap, line 166: technically we can send more than one, so lets remove this check

solvebio/global_search.py
```diff
@@ -148,10 +151,36 @@ def _build_query(self, **kwargs):
...
+            raise AttributeError('Only one entity is supported for search')
+
+        return self._clone(entities=kwargs.items())
```
davecap, line 169: maybe we should cast as a list to be safe `list(kwargs.items())`

krivi95:
@davecap  I'm going to approve this PR, as you're the one who created it and cannot approve it yourself.

I'm not going to merge it yet until everything is double-checked and resolved.

krivi95 (approved):

solvebio/global_search.py
```diff
@@ -26,6 +28,7 @@ def __init__(
             page_size=QueryBase.DEFAULT_PAGE_SIZE,
             result_class=dict,
             debug=False,
+            raw_response=False,
```
davecap, line 31: lets do `raw_results` to be more explicit

davecap:
LGTM!

PR #411, opened 2022-01-19
Global Beacon interface update

krivi95:
## Description

- Return None by default if dataset doesn't contain Global Beacon:

  Updated `Object.get_global_beacon_status` method so it returns `None` by default if Global Beacon doesn't exist.

  ```python
  dataset = Object.get_by_full_path(dataset_full_path)

  # Returns None
  dataset.get_global_beacon_status()

  # Raises exception: API error 404 {"detail": "No Global Beacon for Dataset:12345679"} 
  dataset.get_global_beacon_status(raise_on_disabled=True)
  ```
- Alias methods in the Dataset class for Global Beacons:
  
  Add ability to perform Global Beacon operations with the `Dataset` class:

  ```python
  dataset = Dataset.get_by_full_path(dataset_full_path)
  dataset.get_global_beacon_status()
  dataset.enable_global_beacon()
  dataset.disable_global_beacon()
  ```
 
- Subject method changes to accommodate API changes:
  
  In `subjects()`/`subjects_count()` methods adding additional parameters in the api request `limit=0` and `include_subjects=True`

### DIFF
--- a/solvebio/global_search.py
+++ b/solvebio/global_search.py
@@ -159,7 +159,7 @@ class GlobalSearch(Query):
 
     def execute(self, offset=0, **query):
         # Call superclass method execute
-        Query.execute(self, offset, **query)
+        super(GlobalSearch, self).execute(offset, **query)
 
         # Cast logical objects from response to Object/Vault instances
         if not self._raw_results:
@@ -193,6 +193,6 @@ class GlobalSearch(Query):
         """If entity seaarch is performed returns the list of subjects"""
 
         # Executes a query to get a full API response which contains subjects list
-        self.execute(include_subjects=True, limit=0)
+        self.execute(limit=0)
 
         return self._response.get('subjects_count')
--- a/solvebio/global_search.py
+++ b/solvebio/global_search.py
@@ -107,7 +107,7 @@ class GlobalSearch(Query):
         if filters:
             new._filters += filters
 
-        if limit:
+        if limit is not None:
             new._limit = limit
 
         return new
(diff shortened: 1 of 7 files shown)


solvebio/resource/object.py
```diff
@@ -621,11 +621,19 @@ def disable_global_beacon(self):
...
+    def get_global_beacon_status(self, raise_on_disabled=False):
         """
         Retrieves the Global Beacon status for this object (datasets only).
         """
         if not self.is_dataset:
             raise SolveError("Only dataset objects can be Global Beacons.")
 
-        return self._client.get(self.instance_url() + '/beacon', {})
+        try:
+            response = self._client.get(self.instance_url() + '/beacon', {})
```
davecap, line 632: To make the code a tiny bit cleaner:

```
try:
    return self._client...
except SolveError:
    if raise_o..:
        raise
    return None
```

davecap (approved):
Might want to also add those Dataset object aliases (see my comments about the docs)

solvebio/global_search.py
```diff
@@ -185,14 +185,14 @@ def subjects(self):
...
+        self.execute(include_subjects=True, limit=0)
 
         return self._response.get('subjects')
 
     def subjects_count(self):
         """If entity seaarch is performed returns the list of subjects"""
 
         # Executes a query to get a full API response which contains subjects list
-        self.execute()
+        self.execute(include_subjects=True, limit=0)
```
davecap, line 196: subjects_count is actually always included, so you can set include_subjects=False here

davecap (approved):

solvebio/global_search.py
krivi95: Maybe I'm wrong but haven't you mentioned on the call that currently subjects are always included, but that we're going to change the API so by default it doesn't return subjects all the time, to make it faster?

davecap: Yes, but the subject count field is always returned regardless.

krivi95: Aha ok ok, understood.

solvebio/global_search.py
```diff
@@ -193,6 +193,6 @@ def subjects_count(self):
         """If entity seaarch is performed returns the list of subjects"""
```
davecap, line 193: Docstring is out of date

krivi95: Nice, thanks, missed that one. I'll double-check in the r client as well

PR #414, opened 2022-05-05
add new global search params (vault_scope and entities_match)

davecap:
- Adds missing global search params (vault_scope, entities_match)
- Fix issue with querying objects directly from search results
- Fix issue with test environment dependencies

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -11,7 +11,7 @@ dash_auth==1.4.1
 dash_core_components==1.15.0
 dash_html_components==1.1.2
 dash_renderer==1.9.0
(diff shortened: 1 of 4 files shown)


krivi95 (approved):
👍 

Only to add `vault_scope` and `entities_match` to the [docstring of the contructor](https://github.com/solvebio/solvebio-python/blob/add-new-search-params/solvebio/global_search.py#L38-L48).

davecap:
Also fixes issue with running `result.query()` directly on a constructed Global Search result.

davecap:
Also fixes this: https://github.com/plotly/dash/issues/1992

PR #418, opened 2022-11-04
BDSIM-373: Added recursive data download via sync

nishanthmerwin:
Initial idea... it works but it's missing a few key features:

- `--exclude` option
- delete (if doesn't exist in remote)
- `--dry-run` isn't functional right now
- Better print statements
- tests

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -21,6 +21,7 @@ from solvebio import DatasetImport
 from solvebio import DatasetTemplate
 from solvebio import GlobalSearch
 from solvebio.utils.files import check_gzip_path
+from solvebio.utils.md5sum import md5sum
 from solvebio.errors import SolveError
 from solvebio.errors import NotFoundError
 
@@ -475,11 +476,12 @@ def download(args):
     """
     return _download(args.full_path, args.local_path,
             dry_run=args.dry_run, recursive=args.recursive,
-            excludes=args.exclude, includes=args.include)
+            excludes=args.exclude, includes=args.include,
+            delete=args.delete)
 
 
 def _download(full_path, local_folder_path, dry_run=False, recursive=False,
-        excludes=[], includes=[]):
+        excludes=[], includes=[], delete=False):
     """
     Given a folder or file, download all the files contained
     within it.
@@ -495,7 +497,8 @@ def _download(full_path, local_folder_path, dry_run=False, recursive=False,
                 os.makedirs(local_folder_path)
 
     if recursive:
-        _download_recursive(full_path, local_folder_path, dry_run, excludes, includes)
+        _download_recursive(full_path, local_folder_path, dry_run,
+                excludes, includes, delete)
         return
 
 
@@ -518,7 +521,7 @@ def _download(full_path, local_folder_path, dry_run=False, recursive=False,
 
 
 def _download_recursive(full_path, local_folder_path, dry_run=False,
-        excludes=[], includes=[]):
+        excludes=[], includes=[], delete=False):
 
     if "**" in full_path:
         raise Exception('Paths containing ** are not compatible with the --recursive flag.')
@@ -534,6 +537,10 @@ def _download_recursive(full_path, local_folder_path, dry_run=False,
     for file_obj in results:
         depth = len(file_obj.path.split("/"))
         file_obj.depth = depth
+        # MD5 not retrieved by GlobalSearch so
+        # separate API call is needed
+        file_obj = Object.retrieve(file_obj.id)
+        file_obj.depth = depth
         remote_objects.append(file_obj)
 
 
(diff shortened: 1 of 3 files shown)


solvebio/cli/data.py
```diff
@@ -505,6 +506,113 @@ def _download(full_path, local_folder_path, dry_run=False):
...
+def _get_remote_directory_recursive(full_path, local_folder_path, remote_files=[]):
+    """Queries a SolveBio Vault path and recursively finds all files.
+
+    full_path (str): SolveBio path where data will be downloaded from.
+    local_folder_path (str): Local folder path where data will be downloaded to.
+    remote_files (list): List used to store queried files in recursive search.
+    """
+    objects = Object.all(glob=full_path, limit=1000)
+    for object_ in objects:
+        # TODO: Skip dataset objects for now?
```
davecap, line 526: I think that is fine, just leave a comment and we can also log/print that a dataset (log the ID and the full path) is being skipped

solvebio/cli/data.py
```diff
@@ -505,6 +506,113 @@ def _download(full_path, local_folder_path, dry_run=False):
...
+
+
+def _get_remote_directory_recursive(full_path, local_folder_path, remote_files=[]):
+    """Queries a SolveBio Vault path and recursively finds all files.
+
+    full_path (str): SolveBio path where data will be downloaded from.
+    local_folder_path (str): Local folder path where data will be downloaded to.
+    remote_files (list): List used to store queried files in recursive search.
+    """
+    objects = Object.all(glob=full_path, limit=1000)
```
davecap, line 524: One thing to check is whether this returns everything recursively or just at a single "depth".

It might be worth taking a look at some of the examples in here to see if they would be of any use: https://docs.solvebio.com/vaults/basics/#searching-within-vaults

solvebio/cli/data.py
```diff
@@ -505,6 +506,113 @@ def _download(full_path, local_folder_path, dry_run=False):
...
+        remote_files.append(object_)
+        if object_.is_folder:
+            _get_remote_directory_recursive(
+                    full_path=object_.full_path + "/*",
+                    local_folder_path=object_.local_path,
+                    remote_files=remote_files)
+    return remote_files
+
+
+def _get_local_directory_recursive(local_folder_path):
```
davecap, line 540: This isn't recursive yet or is it?

solvebio/cli/data.py
```diff
@@ -505,6 +506,113 @@ def _download(full_path, local_folder_path, dry_run=False):
...
+    for local_path, file_obj_map in local_file_map.items():
+        print(file_obj_map)
+        remote_file = file_obj_map['remote']
+
+        if remote_file.object_type == "folder":
+            os.makedirs(local_path, exist_ok=True)
+            continue
+
+        if "local" in file_obj_map:
+            if file_obj_map['remote']['md5'] != file_obj_map['local']['md5']:
```
davecap, line 606: It "might" be possible for some remote files to not have md5. In this case, just set the remote md5 to None and assume it is always different (which might already be happening here)

davecap (commented):
Looks good so far

solvebio/cli/data.py
nishanthmerwin: Yeah, so my thinking was that the `sync` method while still using the `glob` search would only allow you to do "flat" searches. This can be accomplished by checking if there's a `**` in the query: https://github.com/solvebio/solvebio-python/pull/418/files#diff-d2e9c9fe10a5684bd8ff375bf4aee2594d5cfef7e9da5980759badcc0ded2b40R562

solvebio/cli/data.py
nishanthmerwin: Python built-in functions handle the recursive search for us!
https://docs.python.org/3/library/os.html#os.walk

I can rename this function though to make that more clear.

solvebio/cli/data.py
nishanthmerwin: Will do. I _think_ right now it would just fail via a KeyError if `md5` didn't exist? I'm not sure if the server actually returns `md5 = null` or just omits the key. I'll take a closer look and adjust accordingly.

nishanthmerwin:
@davecap Thanks for the review. A couple questions:

- Do you like the implementation of having this functionality under a new command called `sync`?
  - Pros: no issues with backward compatibility
  - Cons: Right now sync just implements the download function, not the upload also which is what you typically see with software like aws CLI. Is this ok?
  - Some other options:
    - Implement this as `solvebio download --recursive`. If that flag exists, call this method, otherwise call the existing download method.
    - Add a way to link the upload method to sync so that users can specify either download or upload. e.g. `solvebio sync --download` and `solvebio sync --upload`.

solvebio/cli/data.py
davecap: Look into using GlobalSearch instead of glob/path/query params.

davecap:
Just to confirm, we'll go with `solvebio download --recursive` with the flag to indicate the new code branch.

nishanthmerwin:
> Just to confirm, we'll go with `solvebio download --recursive` with the flag to indicate the new code branch.

This has been implemented.


Updates from today:
- Now can use `--exclude` and `--include` similar to `aws sync`
- Download now uses GlobalSearch instead of `Object.filter(glob=)`. This way a full recursive search can be found in a single API call. Much faster than recursive API calls and also isn't limited to a page size of 1000.
- No longer `solvebio sync`, now it's `solvebio download --recursive`


Things not implemented:
- md5 checking. This had to be removed because GlobalSearch results don't return md5 by default. I'll have to dig into this later, but I don't think that should block this development. Download functions typically overwrite anyways.
- Dry run needs to be checked and implemented thoroughly.

nishanthmerwin:
Ok, so the two main features that are missing are:
- MD5 checking
- Delete local files that are missing.

Can we chat about the use of these features?

solvebio/cli/data.py
```diff
@@ -534,6 +537,10 @@ def _download_recursive(full_path, local_folder_path, dry_run=False,
     for file_obj in results:
         depth = len(file_obj.path.split("/"))
         file_obj.depth = depth
+        # MD5 not retrieved by GlobalSearch so
```
davecap, line 540: only look for md5 if the "object_type" is "file"

nishanthmerwin: Ah right. I realized that we couldn't do MD5 search by "if file > 1MB" because GlobalSearch doesn't return file size either. Will make this change.

davecap: Ahh ok. If you want you can add a feature request ticket to jira for exposing md5 and file size via the base GlobalSearch record schema.

davecap: but this is fine for now I think

nishanthmerwin:
@davecap the last changes should fully support the use case outlined here: https://precisionformedicine.atlassian.net/browse/BDSIM-320

Let me know if there are any changes you suggest. If not, this PR is ready to go IMO.

davecap:
A few notes from testing:

- If you use --recursive where the full path is "folder/*" it doesn't find anything (removing the * or /* works)
- excluding something by name doesnt always work: if I try `--exclude "foldername"` it does not exclude the folder but if I do `--exclude "*foldername"` it works. Similarly if I do `--exclude "foldername/*"` it doesnt seem to work.

I think both of these are related to something you already brought up which is that these globs are based on the full path not relative path. For that reason I think this is probably safe to merge.

davecap:
I also just noticed that download skips empty folders

nishanthmerwin:
- print warning if recursive and ends with "/*"
- Add documentation to docs repo.

nishanthmerwin:
> 



> * Add documentation to docs repo.

image

Added documentation to qbdocs.

davecap (approved):

PR #419, opened 2022-11-17
run black and enable github actions on PRs

davecap:


.github/workflows/python-package.yml
```diff
@@ -1,6 +1,6 @@
 name: SolveBio Python Package
 
-on: push
+on: [push, pull_request]
```
nishanthmerwin, line 3: This runs duplicate pipelines for each change to a pull request.

PR #422, opened 2022-11-23
BDSIM-411: Implement uploads in parallel via CLI

nishanthmerwin:
Ticket: https://precisionformedicine.atlassian.net/browse/BDSIM-411

Current uploads can be really slow if there are thousands of files to upload. Most of this time is IO bound waiting for API requests so we can implement some degree of parallelization. This implementation strategy allows users to specify the number of parallel processes via a CLI option (`--num-processes`).

This implementation _should_ preserve all original functionality by default and will only perform multiprocessing uploads if users specify a flag. Otherwise, uploads will still happen in a multiprocessing.Pool, but with only 1 parallel process.

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -251,18 +251,6 @@ def _create_file_job(args):
         raise e
     except Exception as e:
         return e
-            if dry_run:
-                print(
-                    "[Dry Run] Uploading {} to {}".format(
-                        local_file_path, remote_folder_full_path
-                    )
-                )
-            else:
-                remote_parent = Object.get_by_full_path(
-                    remote_folder_full_path, assert_type="folder"
-                )
-                Object.upload_file(local_file_path, remote_parent.path,
-                                    vault.full_path, archive_folder=archive_folder)
 
 
 def _create_template_from_file(template_file, dry_run=False):
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -11,6 +11,7 @@ import json
 from fnmatch import fnmatch
 from collections import defaultdict
 import multiprocessing
+import signal
 
 import solvebio
 
@@ -97,6 +98,43 @@ def should_exclude(path, exclude_paths, dry_run=False, print_logs=True):
     return False
 
 
+def _check_uploaded_folders(base_remote_path, local_start, all_folders):
+    """Identifies which remote folders already exist so that we can optimize
+    which folders to create in the upload workflow.
+
+    Note, due to the asynchronous nature of GlobalSearch, returned folders
+    may sometimes already exist. A subsequent exact check is required before
+    uploading (as implemented in `_create_folder`).
+
+    Args:
+        base_remote_path: Base remote parent folder full path to upload to.
+        local_start: The name of the base folder to upload.
+        all_folders: A list of remote folders to potentially create in full path format
+    Returns:
+        all_folder_parts (set): A unique set of folder parts to create that do
+            not already exist.
+
+    """
+    upload_root_path, _ = Object.validate_full_path(
+        os.path.join(base_remote_path, local_start)
+    )
+    results = GlobalSearch().filter(path__prefix=upload_root_path, type="folder")
+    remote_folders_existing = set([x.full_path for x in results])
+    all_folder_parts = set()
+    for vault, remote_folder_path in all_folders:
+        subfolders = remote_folder_path.lstrip("/").split("/")
+        parent_folder_path = subfolders[0]
+        # Split each folder into parts and check if these exist
+        # Skip root folder as we don't need to create this
+        for folder in subfolders[1:]:
+            folder_full_path = os.path.join(parent_folder_path, folder)
+            if folder_full_path not in remote_folders_existing:
+                all_folder_parts.add(folder_full_path)
+            parent_folder_path = folder_full_path
+
+    return all_folder_parts
+
+
 def _upload_folder(
     domain,
     vault,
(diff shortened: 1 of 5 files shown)


solvebio/cli/data.py
```diff
@@ -159,10 +162,63 @@ def _upload_folder(
...
+    )
+    results = GlobalSearch().filter(path__prefix=upload_root_path, type="folder")
+    remote_folders_existing = set([x.full_path for x in results])
+    all_folder_parts = set()
+    for vault, remote_folder_path in all_folders:
+        subfolders = remote_folder_path.lstrip("/").split("/")
+        parent_folder_path = ""
+        for folder in subfolders:
+            folder_full_path = os.path.join(parent_folder_path, folder)
+            if folder_full_path not in remote_folders_existing and not folder_full_path.endswith(":"):
```
davecap, line 179: What is the endswith(:) looking for? Can folders have : in them?

solvebio/cli/data.py
```diff
@@ -159,10 +162,63 @@ def _upload_folder(
...
                 )
             else:
-                remote_parent = Object.get_by_full_path(
-                    remote_folder_full_path, assert_type="folder"
-                )
-                Object.upload_file(local_file_path, remote_parent.path, vault.full_path)
+                all_files.append((local_file_path, remote_folder_full_path, vault.full_path))
+
+    # Identify which remote folders already exist
+    upload_root_path, _ = Object.validate_full_path(
```
davecap, line 168: Would it be possible to only call this code block if num processes > 1? just for backwards compatibility?

solvebio/cli/data.py
```diff
@@ -159,10 +162,63 @@ def _upload_folder(
...
+        all_folder_parts = sorted(all_folder_parts, key=lambda x: len(x.split("/")))
+        for folder in all_folder_parts:
+            _create_folder(vault, folder)
+
+        # Create files in parallel
+        pool = multiprocessing.Pool(num_processes)
+        for result in pool.imap_unordered(_create_file_job, all_files):
+            # Check if an exception was raised by the pool worker
+            # which is returned at the end of the tuple
+            if len(result) == 4:
```
davecap, line 196: Does upload_file retry at all in the event of (internal) failure?

solvebio/cli/data.py
```diff
@@ -159,10 +162,63 @@ def _upload_folder(
...
-                    remote_folder_full_path, assert_type="folder"
-                )
-                Object.upload_file(local_file_path, remote_parent.path, vault.full_path)
+                all_files.append((local_file_path, remote_folder_full_path, vault.full_path))
+
+    # Identify which remote folders already exist
+    upload_root_path, _ = Object.validate_full_path(
+        os.path.join(base_remote_path, local_start)
+    )
+    results = GlobalSearch().filter(path__prefix=upload_root_path, type="folder")
```
davecap, line 171: should this block of code be split out into its own function? maybe a few comments to explain what exactly it is doing.

solvebio/cli/data.py
```diff
@@ -159,10 +162,63 @@ def _upload_folder(
...
+    for vault, remote_folder_path in all_folders:
+        subfolders = remote_folder_path.lstrip("/").split("/")
+        parent_folder_path = ""
+        for folder in subfolders:
+            folder_full_path = os.path.join(parent_folder_path, folder)
+            if folder_full_path not in remote_folders_existing and not folder_full_path.endswith(":"):
+                all_folder_parts.add(folder_full_path)
+            parent_folder_path = folder_full_path
+
+    if not dry_run:
```
davecap, line 183: It might be really useful to push the dry_run down into _create_folder and _create_file_job so they print exactly what is happening at that level.

davecap (commented):
Looks good so far

solvebio/cli/data.py
nishanthmerwin: Agreed, it's not a good check. It was just to see if we were at the base folder. I removed this check and instead just started from `for folder in subfolders[1:]`

solvebio/cli/data.py
nishanthmerwin: Absolutely. Can make this change.

solvebio/cli/data.py
nishanthmerwin: Yup! Can do.

solvebio/cli/data.py
nishanthmerwin: Sounds good!

Do you think the GlobalSearch should run with `dry_run`?

solvebio/cli/data.py
nishanthmerwin: Yup, I think under `solvebio.resource.object.Object.upload_file` there's a max_retries set to 5 on the s3 upload URL.
Under `solvebio.client.SolveClient.__init__` all EDP api requests also have a total of 5 retries.


What do you mean by internal failure? Usually retries are triggered due to external failures.

solvebio/cli/data.py
davecap: Since it is a read-only step yes I would say so

solvebio/cli/data.py
davecap: Ok great, I just meant internal within the upload file function itself.

solvebio/cli/data.py
```diff
@@ -97,6 +98,43 @@ def should_exclude(path, exclude_paths, dry_run=False, print_logs=True):
     return False
 
 
+def _check_uploaded_folders(base_remote_path, local_start, all_folders):
```
davecap, line 101: add **kwargs,

set `_client = ...` pass it through validate_full_path(...) and GlobalSearch(client=_client)

nishanthmerwin:
@davecap I tried moving `dry_run` into `_create_folder` however this makes the dry run much slower. Basically, it has to run the check if a folder already exists: `new_obj = Object.get_by_full_path(full_path)` which takes ~0.5s per request. Previously, this was just assumed to be False so we print that we'd create this folder in the `dry_run`.

So now.. if we have ~300 folders to create (as is with the vSIM work), you'd wait about 2 minutes for the dry run to finish. Before, it was instantaneous.

nishanthmerwin:
> I wonder if there is some way to optimize _create_folder based on the information we got from global search?

Definitely. But will be implemented outside of this PR: https://precisionformedicine.atlassian.net/browse/SBP-186?atlOrigin=eyJpIjoiZTY3ZmMwZjJjMTlhNGMzMzk5NTU1OTk4ZDRlMDVjYmEiLCJwIjoiaiJ9

solvebio/cli/data.py
nishanthmerwin: As a general pattern, the idea is that `solvebio.cli` code shouldn't be run outside the CLI context. Anything that we move outside of here (e.g. _create_folder into `Object.create_folder`) will be set to accept `client`.

One of the goals of this ticket is to refactor `create_folders` to incorporate more of this logic: https://precisionformedicine.atlassian.net/browse/SBP-186?atlOrigin=eyJpIjoiZTY3ZmMwZjJjMTlhNGMzMzk5NTU1OTk4ZDRlMDVjYmEiLCJwIjoiaiJ9 so it will be done there.

solvebio/cli/data.py
nishanthmerwin: Still not sure what you mean by this. Is there anything we want to see changed here?

davecap: No worries I think it's all good. Was just wondering if the underlying upload function had retry logic built-in so that we don't need additional logic outside in the even that one of the multiprocessing tasks fails.

davecap (approved):

PR #425, opened 2022-11-29
Allow silent uploads

nishanthmerwin:
In the Makefiles, the following command is often used:

```
LAST_EDC_UPLOAD=$$(solvebio ls $(EDP_VAULT):/$(DEVMODE)/$(PROJECT)/data/source/vSIM_EDC/ | sort | uniq | tail -1 | awk '{print $$2}'; ) ; \
```

Here, if no files are found, the previous behaviour was to print a help text. However, this interferes with the downstream logic. This PR introduces a `--silent` option to this function so that no help texts are printed.

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -720,13 +720,14 @@ def ls(args):
     """
     Given a SolveBio remote path, list the files and folders
     """
-    return _ls(args.full_path)
+    return _ls(args.full_path, silent=args.silent)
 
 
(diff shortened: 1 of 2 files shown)


davecap (approved):

davecap:
@nishanthmerwin should we merge this?

nishanthmerwin:
> @nishanthmerwin should we merge this?

I think we can close this PR since vSIM is able to run without this change.

PR #429, opened 2023-04-19
Skipping vaults from download command

kuzentio:
Skipping Vaults for being downloaded when download recursively.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.21.0
+current_version = 2.22.0
 files = solvebio/version.py
 commit = True
 tag = True
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.22.0
+current_version = 2.21.0
 files = solvebio/version.py
 commit = True
 tag = True
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -645,6 +645,8 @@ def _download_recursive(
         file_obj.depth = depth
         # MD5 not retrieved by GlobalSearch so
         # separate API call is needed
+        if file_obj.class_name() == "Vault":
+            continue
         if file_obj.is_file:
             file_obj = Object.retrieve(file_obj.id)
         file_obj.depth = depth
--- a/solvebio/version.py
+++ b/solvebio/version.py
@@ -3,4 +3,4 @@
 
 # This file should define a variable VERSION which we use as the
 # debugger version number.
-VERSION = '2.21.0'
+VERSION = '2.22.0'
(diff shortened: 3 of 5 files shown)


solvebio/cli/data.py
```diff
@@ -645,6 +645,9 @@ def _download_recursive(
         file_obj.depth = depth
         # MD5 not retrieved by GlobalSearch so
         # separate API call is needed
+        if file_obj.class_name() == "Vault":
+            print("Do not processing Vaults, skipping {}".format(file_obj['vault']))
```
nishanthmerwin, line 649: I don't think we need to print to warn users. We should just silently skip it.

solvebio/version.py
```diff
@@ -3,4 +3,4 @@
 
 # This file should define a variable VERSION which we use as the
 # debugger version number.
-VERSION = '2.21.0'
+VERSION = '2.21.1'
```
nishanthmerwin, line 6: Can you use bumpversion to increment the patch version?

nishanthmerwin:
@kuzentio the code changes look great. Can we add a test case to ensure this is fixed?

I would just add to here: https://github.com/solvebio/solvebio-python/blob/master/solvebio/test/test_shortcuts.py#L522
Add another entry with [fake_vault_create](https://github.com/solvebio/solvebio-python/blob/6d71d448dcb5389071cc34531cf71a7743aa6eae/solvebio/test/client_mocks.py#L239)

requirements-dev.txt
```diff
@@ -17,3 +17,4 @@ pyyaml==5.3.1
 click==7.1.2
 ruamel.yaml==0.16.12
 pytest
+requests
```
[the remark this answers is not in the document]
nishanthmerwin: https://github.com/solvebio/solvebio-python/blob/master/setup.py#L28

How did you install the package? Pretty sure it already installs requests (see link above) if you use `pip install -e .`

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.21.0
+current_version = 2.22.0
```
davecap, line 2: we should bump version separately

davecap:
@kuzentio @nishanthmerwin I'm honestly not sure what changed... maybe github changed the way their action runners work because they used to just run on github. Now it is saying we don't have the ability to use github-hosted runners.

nishanthmerwin:
> @kuzentio @nishanthmerwin I'm honestly not sure what changed... maybe github changed the way their action runners work because they used to just run on github. Now it is saying we don't have the ability to use github-hosted runners.

Found the error: https://stackoverflow.com/a/70968478
Looks like we just need to bump up ubuntu. They discontinued it on April 1st!

davecap:
@nishanthmerwin nice!!!

nishanthmerwin:
Note that a _lot_ of things start breaking in ubuntu 22. Pretty much all the python 2 stuff stops working. I don't think github has announced when they'll start taking it down, but expect we have ~2 years left for this to work before we have to stop supporting python 2.

nishanthmerwin (approved):
Removed the version bump within this PR. Can do this afterwards.

PR #430, opened 2023-05-02
[EDPDEV-316] Don't re-query for md5 when using GlobalSearch

nishanthmerwin:
Previously, when downloading files recursively, we used GlobalSearch to first find the files to download and then subsequently have to make a file object request to retrieve MD5 information. GlobalSearch was modified in the API to return md5 information, so we no longer need to make a second query.

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -647,7 +647,7 @@ def _download_recursive(
         # separate API call is needed
         if file_obj.class_name() == "Vault":
             continue
-        if file_obj.is_file:
+        if file_obj.is_file and not file_obj.get('md5'):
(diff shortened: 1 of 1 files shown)


solvebio/cli/data.py
```diff
@@ -688,7 +688,8 @@ def _download_recursive(
             "{}Downloading file {}".format("[Dry run] " if dry_run else "", local_path)
         )
         if not dry_run:
-            os.makedirs(parent_dir, exist_ok=True)
```
nishanthmerwin, line 691: os.makedirs doesn't have the parameter `exist_ok` in python 2.7, so it would fail. This only works in python >3

damnjanovictanja (approved):

PR #431, opened 2023-05-09
EDPDEV-206 remove checking already exists folders

kuzentio:
I have mange to reproduce it for my env. 
For me removing this check helps, but I'm still not sure why. On API it fails when try to fetching index, maybe some authorization is missing, just thinking.


nishanthmerwin (approved):
Looks good to me!

PR #433, opened 2023-06-14
Draft: Added Credentials Manager

oboforty:
- Added a class that handles finding credentials. This is only used for unit tests, but my intention for the future is to use it for solvebio cli, solvebio.__main__.py, and for solvebio.SolveClient, solvebio.login in case someone uses this as an SDK.

Credentials are discovered in the following priority:
1. Direct parameters (when using solvebio.SolveClient() or solvebio.login() from the SDK)

    This way is naturally NOT supported for the following setups:
    - Solvebio python entry point (solvebio as a global command)
    - Running solvebio as package (__main__.py)
    - Solvebio unit tests
2. Environment variables
    - SOLVEBIO_API_HOST
    - SOLVEBIO_ACCESS_TOKEN
    - SOLVEBIO_API_KEY
3. System credentials file - can be found in:
    - Nix:      ~/.solvebio/credentials
    - Windows:  %USERPROFILE%/_solvebio/credentials


oboforty:
Let's leave this PR hanging for the time being and return to it after the backend tech debts have been dealt with.

This PR will be dedicated for the Creds manager, that follows a similar flow to boto3 in finding the credentials (which was probably the intention already originally for solvebio client). The nosetest & py3.10 fix is added to a separate PR.

davecap:
FYI we have a new way of handling credentials in the new beta R client, so we can revisit this and re-align it to that new convention.

PR #434, opened 2023-06-19
EDPDEV-567: remove nosetests & fixed python 2.7 tests

oboforty:


davecap:
@PolinaBevad would you like to test this branch on Python 3.10? It should be working now.

davecap (approved):

solvebio/test/test_object.py
```diff
@@ -18,6 +20,7 @@ def get_uuid_str():
 
 class ObjectTests(SolveBioTestCase):
 
+    @skip
```
davecap, line 23: @oboforty can we remove skips?

PR #436, opened 2023-08-10
Add recursive option for ls command

davecap:


solvebio/cli/data.py
```diff
@@ -743,6 +743,8 @@ def _ls(full_path):
                 file_.last_modified, file_.object_type.ljust(8), file_.full_path
             )
         )
+        if recursive and file_.object_type == "folder":
+            _ls(file_.full_path + "/*", recursive=True)
 
     if len(files) == 0:
```
nishanthmerwin, line 749: This logic needs to be modified if this is now a recursive function. Currently, if there's any empty directory in the descendents, it prints out the "No files found..." message.

davecap: Ah, makes sense, thanks

davecap: Pushed a change to handle this

nishanthmerwin (approved):
LGTM!

davecap:
@nishanthmerwin can you give the latest a try when you have a few mins

nishanthmerwin:
Tried it on a big folder:

```
❯ solvebio ls "admin-qb-int-dev:Nishanth's Test Vault:/source" --recursive | wc -l
1468
❯ time !!
❯ time solvebio ls "admin-qb-int-dev:Nishanth's Test Vault:/source" --recursive | wc -l
1468
solvebio ls "admin-qb-int-dev:Nishanth's Test Vault:/source" --recursive  0.34s user 0.01s system 2% cpu 16.094 total
wc -l  0.00s user 0.00s system 0% cpu 16.094 total
```

It's not the fastest, but it does the job. Looked through the output and didn't see anything too off.

PR #437, opened 2023-09-11
[EDPDEV-1000] Fix upload bug

nishanthmerwin:
The issue was related to the session object not being thread-safe. Previously, a single session object was shared across all processes making requests leading to really strange errors (e.g. ECONNRESET). 

This change creates a new session object for each process. I have tested this by running the following command to upload ~1000 files to an EDP instance:

```
solvebio upload --full-path "edp:Nishanth's Test Vault:/folder" ./source --num-processes 20 --create-full-path
```

Previously, this would fail immediately after creating the folders. Now, this entire process succeeds.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,3 +16,4 @@ pyyaml==5.3.1
 click==7.1.2
 ruamel.yaml==0.16.12
 pytest
+brotli<=1.0.9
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -16,3 +16,4 @@ pyyaml==5.3.1
 click==7.1.2
 ruamel.yaml==0.16.12
 pytest
+brotli<=1.0.9
```
nishanthmerwin, line 19: Pipelines were failing in py27: https://github.com/solvebio/solvebio-python/actions/runs/6150337104/job/16688071606

It was due to this breaking change: https://github.com/google/brotli/issues/1074

davecap (approved):

PR #440, opened 2023-10-05
EDPDEV-1049 fix issue with recursive uploads not using correct global credentials

davecap:


solvebio/cli/data.py
```diff
@@ -200,12 +204,13 @@ def _create_file_job(args):
         None or Exception if exception is raised.
```
nishanthmerwin, line 204: Minor: Update args here with the new set. This repo doesn't have a ton of docstrings, but because of the way args have to be sent into the method here, we should keep the docs updated.

PR #443, opened 2023-10-25
[EDPDEV-1075] Update env var token names

nishanthmerwin:
@davecap the new R client expects to find access tokens and credentials at `EDP_ACCESS_TOKEN` and `EDP_API_HOST` respectively. What do you think about adding this as a backup for this client?

This was also causing our vsim docker image to fail: https://precisionformedicine.atlassian.net/browse/EDPDEV-1075?focusedCommentId=155033


nishanthmerwin:
See R documentation here: https://quartzbio.github.io/quartzbio.edp/reference/connect.html

davecap (approved):

PR #445, opened 2023-10-26
EDPDEV 938 Parallel Download

nrasulicpfm:
This pull request implements parallel downloading using a ThreadPoolExecutor.

It is built on top of:
https://github.com/solvebio/solvebio-python/pull/439

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -27,6 +27,7 @@ from solvebio.utils.md5sum import md5sum
 from solvebio.errors import SolveError
 from solvebio.errors import NotFoundError
 from solvebio.client import SolveClient
+from solvebio.client import client as global_client
 
 
 def should_exclude(path, exclude_paths, dry_run=False, print_logs=True):
@@ -150,11 +151,14 @@ def _upload_folder(
             all_folders.append((vault, remote_path))
 
         # Upload the files that do not yet exist on the remote
+        # Pass global client auth parameters to each worker to avoid using default globals
+        client_auth = (global_client._host, global_client._auth.token, global_client._auth.token_type)
         for f in files:
             local_file_path = os.path.join(abs_local_parent_path, f)
             if should_exclude(local_file_path, exclude_paths, dry_run=dry_run):
                 continue
-            all_files.append((local_file_path, remote_folder_full_path, vault.full_path, dry_run, archive_folder))
+            all_files.append((local_file_path, remote_folder_full_path, vault.full_path,
+                              dry_run, archive_folder, client_auth))
 
     if num_processes > 1:
         # Only perform optimization if parallelization is requested by the user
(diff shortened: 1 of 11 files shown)


solvebio/cli/data.py
```diff
@@ -633,26 +640,51 @@ def _download(
             )
         )
 
+    if not follow_shortcuts:
```
nishanthmerwin, line 643: Should we inform users e.g. "5 shortcuts were found but not downloaded, use `--follow-shortcuts` flag to download."?

solvebio/cli/data.py
```diff
@@ -735,6 +795,70 @@ def _download_recursive(
...
+    for shortcut in shortcuts:
+        # Global search doesn't return target in the response.
+        # Directly accessing target will always return None.
+        # Call Object.retrieve() to get the full object.
+        target_object = Object.retrieve(shortcut.id).get_target()
+        if not target_object:
+            print("Couldn't find target object for shortcut: {}".format(shortcut.full_path))
+            continue
+        if target_object.is_folder:
+            # todo # implement circular shortcut avoiding
```
nishanthmerwin, line 836: Is this needed since the backend already prevents circular shortcuts?

nishanthmerwin: It looks like `visited_folders` isn't used at all?

solvebio/cli/data.py
nishanthmerwin: E.g. when I tried to download from a vault with only a shortcuts, I get the following message:

```
❯ solvebio download --recursive "admin-qb-int-dev:nish-2-vault:/" .
Found 0 files to download.
No files found on path.
```

Since we have the information of how many shortcuts there are, we should at least let the user know.

nishanthmerwin:
Trying to download from this vault: https://admin-qb-int-dev.edp.aws.quartz.bio/vaults?vault=326

I used this command and got an error:

```
❯ solvebio download --recursive --follow-shortcuts "admin-qb-int-dev:nish-2-vault:/" .
Traceback (most recent call last):
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/resource/solveobject.py", line 59, in __getattr__
    return self[k]
KeyError: 'is_folder'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/nishanth_merwin/miniconda3/envs/solvebio_client/bin/solvebio", line 33, in <module>
    sys.exit(load_entry_point('solvebio', 'console_scripts', 'solvebio')())
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/cli/main.py", line 505, in main
    return args.func(args)
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/cli/data.py", line 580, in download
    return _download(
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/cli/data.py", line 617, in _download
    _download_recursive(
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/cli/data.py", line 679, in _download_recursive
    results = list(_resolve_shortcuts_and_get_files(full_path=full_path,
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/cli/data.py", line 835, in _resolve_shortcuts_and_get_files
    if target_object.is_folder:
  File "/home/nishanth_merwin/projects/edp-python-client/solvebio-python/build/lib/solvebio/resource/solveobject.py", line 61, in __getattr__
    raise AttributeError(*err.args)
AttributeError: is_folder
```

solvebio/cli/data.py
nrasulicpfm: Yes visited folders isn't used.
I tried implementing some circular shortcut detection with it but it didn't work out.
Left it in there to still keep track of the folders, but it can be removed.

This is one of the circular scenarios I thought of:
image

Something like this isn't handled on the backend.
One way I can propose to guard against this is to implement a graph and use graph traversal to detect potential circular shortcuts.

solvebio/cli/data.py
nrasulicpfm: You're right. I'll add this to the base 'recursive_download' branch and then merge into this one, once finalized there.

solvebio/cli/data.py
davecap: I would add this as a future enhancement

davecap (approved):

PR #447, opened 2023-12-06
Add Connect to Client Retry Configuration

nrasulicpfm:
This PR sets the connect parameter to the session Retry.
As per the documentation this should enable retries on dropped connections.
Linked ticket:
https://precisionformedicine.atlassian.net/browse/EDPDEV-1175

And added an explicit list of allowed methods.
This includes non-idempotent request - however imo if the retries only occur on 502 503 504 or a not established connection this should not cause an issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ For all other requests, please [email SolveBio Support](mailto:support@solvebio.
 Configuring the Client
 -------
 
-The Solvebio python client can be configured by setting system environment variables.
+The SolveBio python client can be configured by setting system environment variables.
 Supported environment variables are:
 
 `SOLVEBIO_API_HOST`     
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ The Solvebio python client can be configured by setting system environment varia
 Supported environment variables are:
 
 `SOLVEBIO_API_HOST`     
-- The URL of the target API backend.
-                          If not specified the value from the local credentials file will be used.
+- The URL of the target API backend. 
+If not specified the value from the local credentials file will be used.
 
 `SOLVEBIO_ACCESS_TOKEN` 
 - The OAuth2 access token for authenticating with the API.
@@ -100,7 +100,7 @@ The lookup order for credentials is:
 
 `SOLVEBIO_LOGLEVEL` 
 - The log level at which to log messages.
-                          If not specified the default log level will be WARN.
+If not specified the default log level will be WARN.
 
 `SOLVEBIO_LOGFILE`        
 - The file in which to write log messages. 
--- a/README.md
+++ b/README.md
@@ -75,3 +75,41 @@ Developer documentation is available at [docs.solvebio.com](https://docs.solvebi
 If you experience problems with this package, please [create a GitHub Issue](https://github.com/solvebio/solvebio-python/issues).
 
 For all other requests, please [email SolveBio Support](mailto:support@solvebio.com).
+
+
+Configuring the Client
+-------
+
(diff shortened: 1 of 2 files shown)


solvebio/client.py
```diff
@@ -107,12 +107,23 @@ def __init__(self, host=None, token=None, token_type='Token',
         # intermittent connection errors.
         retries = Retry(
             total=5,
+            connect=5,
```
davecap, line 110: Does this have an impact if we already set `total`?

Can we increase the backoff factor to `2`?

solvebio/client.py
```diff
@@ -107,12 +107,23 @@ def __init__(self, host=None, token=None, token_type='Token',
...
             backoff_factor=0.1,
             status_forcelist=[
                 codes.bad_gateway,
                 codes.service_unavailable,
                 codes.gateway_timeout
+            ],
+            allowed_methods=frozenset([
```
davecap, line 117: Any thoughts on using an environment variable to include non-idempotent methods to the list, or some kind of "aggressive retry mode" for people going through proxies?

`SOLVEBIO_RETRY_ALL=1` or something to enable PUT/POST/PATCH/DELETE?
We can enable HEAD/GET/OPTIONS/TRACE by default

nrasulicpfm: yes we can make it like that - I'll get on it

nrasulicpfm:
addressed the PR comments

solvebio/client.py
```diff
@@ -103,26 +104,39 @@ def __init__(self, host=None, token=None, token_type='Token',
         }
         self.set_user_agent()
 
+        solvebio_retry_all = os.environ.get('SOLVEBIO_RETRY_ALL')
```
davecap, line 107: I would just wrap this in a `bool()` rather than require `1`

solvebio/client.py
```diff
@@ -103,26 +104,39 @@ def __init__(self, host=None, token=None, token_type='Token',
...
+        if solvebio_retry_all == '1':
+            logger.info("Retries enabled for all operations.")
```
davecap, line 109: Retries enabled for all API requests

solvebio/client.py
```diff
@@ -103,26 +104,39 @@ def __init__(self, host=None, token=None, token_type='Token',
...
+                "GET",
+                "PUT",
+                "POST",
+                "PATCH",
+                "DELETE",
+                "OPTIONS",
+                "TRACE"
+            ])
+        else:
+            logger.info("Retries enabled only for idempotent operations.")
```
davecap, line 121: Retries enabled for read-only API requests

davecap (changes requested):
Minor changes requested

solvebio/client.py
nrasulicpfm: 👍

solvebio/client.py
nrasulicpfm: 👍

nrasulicpfm:
Addressed all of the comments :)

davecap (approved):

README.md
```diff
@@ -75,3 +75,41 @@ Developer documentation is available at [docs.solvebio.com](https://docs.solvebi
 If you experience problems with this package, please [create a GitHub Issue](https://github.com/solvebio/solvebio-python/issues).
 
 For all other requests, please [email SolveBio Support](mailto:support@solvebio.com).
+
+
+Configuring the Client
+-------
+
+The Solvebio python client can be configured by setting system environment variables.
```
davecap, line 83: SolveBio Python client

PR #448, opened 2023-12-27
Add --follow-shortcuts to ls Command

nrasulicpfm:
This pull request adds support for shortcuts when using the `ls` command.
The command would add `--follow-shortcuts` flag similar to `upload` and `download`

### DIFF
--- a/solvebio/cli/data.py
+++ b/solvebio/cli/data.py
@@ -942,7 +942,9 @@ def ls(args):
               "Try the --recursive flag instead.")
         return False
 
-    files = _ls(args.full_path, recursive=args.recursive)
+    follow_shortcuts = True if args.follow_shortcuts else False
+
+    files = _ls(args.full_path, recursive=args.recursive, follow_shortcuts=follow_shortcuts)
 
     if len(files) == 0:
         print(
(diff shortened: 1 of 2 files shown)


solvebio/cli/data.py
```diff
@@ -942,7 +942,9 @@ def ls(args):
               "Try the --recursive flag instead.")
         return False
 
-    files = _ls(args.full_path, recursive=args.recursive)
+    follow_shortcuts = True if args.follow_shortcuts else False
```
davecap, line 945: why do we need this if we can just use `args.follow_shortcuts`

solvebio/cli/data.py
```diff
@@ -954,17 +956,38 @@ def ls(args):
...
+                    "{}  {}  {}  from shortcut: {}".format(
+                        resolved_file.last_modified,
+                        resolved_file.object_type.ljust(8),
+                        resolved_file.full_path.ljust(50),
+                        shortcut
+                    )
+                )
+            except NotFoundError:
+                print(
+                    "shortcut target not found!{}from shortcut: {}".format(
```
davecap, line 977: can we rephrase it to indicate that it may not be accessible to the current user?
`Shortcut {} could not be resolved: the target may have been deleted or you may not have permission to access it`

davecap (changes requested):
Minor changes requested

solvebio/cli/data.py
nrasulicpfm: Addressed these comments a while ago - forgot to follow up here in the thread sorry

solvebio/cli/data.py
nrasulicpfm: Addressed these comments a while ago - forgot to follow up here in the thread sorry

davecap (approved):

davecap:
@nrasulicpfm can we merge this?

PR #450, opened 2024-03-05
EDPDEV-1195 Query Fail Due to Memory Error

rakesh-precisionformedicine:
- Add len, repr and next methods to handle it

### DIFF
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -286,10 +286,16 @@ class QueryBase(object):
     def __repr__(self):
         # Check that Query/QueryFile object does not have any previous errors
         # otherwise, raise the error.
+        total_count = 0
+        try:
+            total_count = len(self)
+        except TypeError:
+            total_count = float("inf")
+
         if self._error:
             raise self._error
 
-        if len(self) == 0:
+        if total_count == 0:
             return 'Query returned 0 results.'
 
         placeholder = 'many more results (total unknown)'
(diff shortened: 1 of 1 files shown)


nrasulicpfm (approved):

solvebio/query.py
```diff
@@ -1220,3 +1220,26 @@ def fields(self):
...
+    def __len__(self):
+        base_len = super(QueryFile, self).__len__()
+        if base_len == float('inf'):
+            raise TypeError(
+                'Unable to determine the number of records for the '
+                'current file-based query'
+            )
+        return base_len
+
+    def __repr__(self):
```
davecap, line 1233: rather than make a new `__repr__` please modify the superclass to be like this:

```
def __repr__(self):
    ...
    try:
        total_count = len(self)
    except TypeError:
        total_count = float('inf')
    
    if total_count == 0:
        return 'Query returned 0 results.'

    ...
```

solvebio/query.py
```diff
@@ -1220,3 +1220,26 @@ def fields(self):
...
+        try:
+            super(QueryFile, self).__repr__()
+        except TypeError:
+            return 'Query returned 0 results.'
+        else:
+            return super(QueryFile, self).__repr__()
+
+    def __next__(self):
```
davecap, line 1241: need to modify the superclass `next()` method instead of this...

From the jira ticket:
> if len(self) raises TypeError, don’t raise StopIteration since we are not yet at the end

In `next()` as it is currently defined, you will see this:

```
# len(self) returns `min(limit, total)` results
if not _is_join and self._cursor == len(self):
    raise StopIteration
```

With the new code, `len(self)` might raise TypeError.. so handle that case here but just continue normally if it does return TypeError...

```
try:
    if not _is_join and self._cursor == len(self):
        raise StopIteration
except TypeError
    # len(self) is unknown so just continue normally
    continue
```

nrasulicpfm:
Closing this PR as it has been handled here
https://github.com/solvebio/solvebio-python/pull/453

PR #453, opened 2024-05-16
EDPDEV-1195 Handle Big Objects

nrasulicpfm:
This PR handles big objects by handling len(self) differently and raising an error if the exact number of records is unkown.
Linked ticket:
https://precisionformedicine.atlassian.net/browse/EDPDEV-1195

Implementation details from ticket comments:

the [QueryFile](https://github.com/solvebio/solvebio-python/blob/master/solvebio/query.py#L1056) class currently sets the default _limit to INF which is a huge number. Instead of this, we can take a different approach which is to raise a TypeError from len() if we don’t know the true number of records

Many types of object “have no len” such as files and things, so this is technically a standard approach

The true max limit calculation happens in def __len__ which is not yet defined in the QueryFile subclass but we can define it there

Change QueryFile default _limit to be float('inf') same as for class Query for internal consistency

Subclass QueryFile.__len__ such that it attempts to determine the accurate length as it does now by considering _limit and total, but raises TypeError("Unable to determine the number of records") if self._limit == float('inf')… this is important because QueryFile.__len__ cannot return anything other than an integer.
Anywhere else in the class where len(self) is called, handle it a bit more carefully

in __repr__, if len(self) raises TypeError, assume the length is float('inf') and continue

in next(),  if len(self) raises TypeError, don’t raise StopIteration since we are not yet at the end

### DIFF
--- a/solvebio/query.py
+++ b/solvebio/query.py
@@ -286,10 +286,16 @@ class QueryBase(object):
     def __repr__(self):
         # Check that Query/QueryFile object does not have any previous errors
         # otherwise, raise the error.
+        total_count = 0
+        try:
+            total_count = len(self)
+        except TypeError:
+            total_count = float("inf")
+
         if self._error:
             raise self._error
 
-        if len(self) == 0:
+        if total_count == 0:
             return 'Query returned 0 results.'
 
         placeholder = 'many more results (total unknown)'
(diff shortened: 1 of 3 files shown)


Rob-precisionformedicine (approved):

Rob-precisionformedicine (approved):

davecap:
LGTM

davecap (approved):

PR #454, opened 2024-05-22
Fix EDPDEV-1696: don't verify SSL for internal connections

davecap:


nrasulicpfm (approved):

davecap:
FYI This change can safely be reverted

PR #465, opened 2024-08-08
Support transparent redirection between solvebio.com and quartz.bio

davecap:


nishanthmerwin:
Implementation looks good to me! I would add a unit test to ensure this works as expected (I assume you just tried it out locally).

nishanthmerwin (approved):
LGTM!

PR #493, opened 2025-08-13
Support for multipart file upload.

rajasimon-precisionformedicine:
File larger than 5Gb failed because of the technical limitation but with the multi part upload feature the files are upload in chunks.


solvebio/resource/object.py
```diff
@@ -442,79 +443,88 @@ def upload_file(cls, local_path, remote_path, vault_full_path, **kwargs):
         mimetype = mime_tuple[1] if mime_tuple[1] else mime_tuple[0]
         # Get file size
         size = os.path.getsize(local_path)
-        if size == 0:
```
damnjanovictanja, line 445: Why did you remove this part? We should skip the empty files.

